PR #3, opened 2025-04-30
fix(ci): No longer depend on secret for CI checkout

cleithner-comcast:
We're public now - we don't need it.


kfundecmcsa (approved):

tleacmcsa (approved):

sbarre01 (commented):
looks good to me

PR #4, opened 2025-04-30
build: add support for rdk yocto builds

kfundecmcsa:
In order to ensure compatibility with the rdk Yocto build environment:

- exclude cmocka as a required dependency if testing is not required
- dynamically resolve include paths for matter and otbr
- reintroduce CMAKE_FIND_ROOT_PATH to `find_library` commands

Refs: BARTON-232


config/cmake/platforms/dev/linux.cmake
```diff
@@ -30,6 +30,8 @@ set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Build type")
 set(BDS_BUILD_WITH_SSP ON CACHE BOOL "Build with SSP")
 set(BDS_BUILD_WITH_ASAN ON CACHE BOOL "Build with ASAN")
 
+set(CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}/matter-install" CACHE PATH "Path to Matter installation")
```
cleithner-comcast, line 33: nit: probably I would throw this at the top and not in between barton public options.

cleithner-comcast (approved):

tleacmcsa (approved):

PR #5, opened 2025-05-01
build!: Refactor library name

cleithner-comcast:
This commit refactors the library targets of barton to bDeviceService (and bDeviceServiceStatic). Additionally, we don't attempt to install bDeviceServiceStatic as it doesn't make sense to install both the shared object and the static library. Clients relying on the static library should reference the cmake target directly or install it out of band.

### DIFF
--- a/api/c/CMakeLists.txt
+++ b/api/c/CMakeLists.txt
@@ -27,16 +27,16 @@ project(brtn-ds-private)
 
 file(GLOB_RECURSE BRTN_DS_PRIVATE_API_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/private/*.c)
 
-add_library(brtnDeviceServicePrivateAPI STATIC ${BRTN_DS_PRIVATE_API_SOURCES})
+add_library(bDeviceServicePrivateAPI STATIC ${BRTN_DS_PRIVATE_API_SOURCES})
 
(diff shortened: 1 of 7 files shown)


cleithner-comcast:
I'm not married to the name, I just thought I'd throw out something less verbose. If we still like "brtnDeviceService" I'm happy to sed it back to that.

The main thing here is to get rid of the "Shared" suffix and not install the static library. Just threw out a new name to get some opinions.

tleacmcsa (commented):
I dunno about 'device service' in the name.  This stuff is now Barton.  How about BartonCore?  libBartonCore, libBartonCorePrivateAPI, etc.

tleacmcsa:
> > I dunno about 'device service' in the name. This stuff is now Barton. How about BartonCore? libBartonCore, libBartonCorePrivateAPI, etc.
> 
> I'd be cool with libBartonCore. It is worth pointing out that our public api is namespaced b_device_service. Maybe that's okay?

Dang, thats right.  Maybe we should rethink all of this naming.  It should be consistent and clear and now is the time to change it, but we would have to do very soon.

kfundecmcsa:
>Dang, thats right. Maybe we should rethink all of this naming. It should be consistent and clear and now is the time to change it, but we would have to do very soon.

This sounds interesting to me

PR #6, opened 2025-05-01
chore: Add .gitmessage on hooks/install.sh

cleithner-comcast:


hooks/install.sh
```diff
@@ -40,3 +40,6 @@ for hook in "${HOOKS_SUPPORTED[@]}"; do
     echo "Installing ${hook} hook."
     cp ${MY_DIR}/${hook} ${HOOKS_DIR}/${hook}
 done
+
+# Additionally, install the .gitmessage
+git config --local commit.template .gitmessage
```
kfundecmcsa, line 45: Probably best to prove the absolute path to the template: `$TOP_DIR/.gitmessage`

kfundecmcsa (changes requested):

hooks/install.sh
cleithner-comcast: Aha, you would think but it appears that the expected path is relative to the git repo root directory.

kfundecmcsa: ha weird! sounds good then

kfundecmcsa (approved):

PR #7, opened 2025-05-02
feat(ref): Set the wifi creds from ref app's cli

Lakshmi97-velampati:
Adds the command to set the wifi credentials from reference app's cli.

Refs: BARTON-240

### DIFF
--- a/reference/src/reference-network-credentials-provider.c
+++ b/reference/src/reference-network-credentials-provider.c
@@ -27,11 +27,16 @@
 
 #include "provider/device-service-network-credentials-provider.h"
 #include "reference-network-credentials-provider.h"
+#include <icConcurrent/threadUtils.h>
 
 struct _BReferenceNetworkCredentialsProvider
 {
     GObject parent_instance;
 };
+
+static pthread_mutex_t mtx = PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP;
+static gchar *network_ssid = NULL;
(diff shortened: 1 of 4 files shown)


reference/src/reference-network-credentials-provider.c
```diff
@@ -27,12 +27,17 @@
...
 #include "provider/device-service-network-credentials-provider.h"
 #include "reference-network-credentials-provider.h"
+#include <icConcurrent/threadUtils.h>
 
 struct _BReferenceNetworkCredentialsProvider
 {
     GObject parent_instance;
 };
 
+static pthread_mutex_t mtx = PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP;
```
cleithner-comcast, line 37: Please avoid using non-descript variable names like mtx. The variable name should describe what it is. In the case of mutexes, it should describe the data it is guarding. Something like "network_creds_mtx" would work, or something similar.

cleithner-comcast (changes requested):
Looks very close to me.

reference/src/reference-network-credentials-provider.c
Lakshmi97-velampati: Updated

cleithner-comcast (approved):

tleacmcsa (approved):

mkkoch (approved):

PR #8, opened 2025-05-05
refactor(core): Use GLib containers in subsystemManager

ganeshInduri434:
This change replaces the use of `icLinkedList` and `icHashMap` with `GList` and `GHashTable`.

Refs: BARTON-245

### DIFF
--- a/core/src/deviceService.c
+++ b/core/src/deviceService.c
@@ -4323,16 +4323,16 @@ DeviceServiceStatus *deviceServiceGetStatus(void)
 
     result->subsystemsJsonStatus = hashMapCreate();
 
-    scoped_icLinkedListGeneric *subsystems = subsystemManagerGetRegisteredSubsystems();
-    sbIcLinkedListIterator *subsystemsIt = linkedListIteratorCreate(subsystems);
-    while (linkedListIteratorHasNext(subsystemsIt))
+    GList *subsystems = subsystemManagerGetRegisteredSubsystems();
+    for (GList *subsystemsIt = subsystems; subsystemsIt != NULL; subsystemsIt = subsystemsIt->next)
     {
-        char *subsystemName = linkedListIteratorGetNext(subsystemsIt);
+        char *subsystemName = (char *) subsystemsIt->data;
         hashMapPut(result->subsystemsJsonStatus,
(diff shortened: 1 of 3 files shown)


core/src/deviceService.c
```diff
@@ -4323,16 +4323,16 @@ DeviceServiceStatus *deviceServiceGetStatus(void)
 
     result->subsystemsJsonStatus = hashMapCreate();
 
-    scoped_icLinkedListGeneric *subsystems = subsystemManagerGetRegisteredSubsystems();
-    sbIcLinkedListIterator *subsystemsIt = linkedListIteratorCreate(subsystems);
-    while (linkedListIteratorHasNext(subsystemsIt))
+    GList *subsystems = subsystemManagerGetRegisteredSubsystems();
```
tleacmcsa, line 4326: Please use g_auto style cleanup.  Especially if refactoring another auto cleanup method (scoped_*)

tleacmcsa (commented):
Refactoring is a perfect time to first add tests (unit in this case). You can then be sure that the tests are good before the refactoring, and that after refactoring it works as well.

core/src/subsystemManager.c
```diff
@@ -61,34 +61,32 @@ typedef struct
...
+static GHashTable *subsystems;
 static bool allDriversStarted = false;
 static subsystemManagerReadyForDevicesFunc readyForDevicesCB = NULL;
 static void checkSubsystemForMigration(SubsystemRegistration *registration);
 
 static void subsystemRegistrationDestroy(SubsystemRegistration *reg)
 {
     pthread_mutex_destroy(&reg->mtx);
-    free(reg);
+    g_free(reg);
```
cleithner-comcast, line 85: For the future, no need to swap frees to g_frees. Technically glib's allocator/deallocator implementations will (mostly) use platform-specific malloc/free but we're not too concerned with an extremely high level of portability - free/malloc/calloc are perfectly fine for non-glib types.

cleithner-comcast:
This looks like it should maybe also be in draft?

core/src/deviceService.c
tleacmcsa: not done.

PR #11, opened 2025-05-12
build: Fix Adding CMake Options to build.sh

cleithner-comcast:
There was a regression where adding custom CMake options to build.sh would simply spit out the help message. This commit fixes build.sh to allow you to padd custom CMake options in addition to the supported build options.


build.sh
```diff
@@ -61,9 +62,9 @@ while [[ $# -gt 0 ]]; do
         shift
         ;;
     *)
-        echo "Unknown option: $1"
-        help
-        exit 1
+        echo "Appending CMake args $1"
+        CMAKE_ARGS=$1
```
mkkoch, line 66: This doesn't seem like it appends, but will just take whatever the last option was

mkkoch (changes requested):

build.sh
cleithner-comcast: Oh excellent point. If I had tried with multiple options, I would have discovered this. d'oh!

tleacmcsa (dismissed):

kfundecmcsa (dismissed):

build.sh
```diff
@@ -61,9 +62,9 @@ while [[ $# -gt 0 ]]; do
         shift
         ;;
     *)
-        echo "Unknown option: $1"
-        help
-        exit 1
+        echo "Appending CMake args $@"
+        CMAKE_ARGS=$@
```
mkkoch, line 66: This should probably be CMAKE_ARGS="$@" otherwise it won't properly preserve quoted arguments. See https://stackoverflow.com/questions/4824590/propagate-all-arguments-in-a-bash-shell-script

build.sh
```diff
@@ -91,7 +92,7 @@ fi
 
 pushd ${MY_DIR}
 
-cmake -B ${BUILD_DIR} ${CMAKE_CACHE_OPTION} $@
+cmake -B ${BUILD_DIR} ${CMAKE_CACHE_OPTION} $CMAKE_ARGS
```
mkkoch, line 95: Again here, "$CMAKE_ARGS"

mkkoch (changes requested):

build.sh
cleithner-comcast: Good catch, thanks.

tleacmcsa (approved):

PR #13, opened 2025-05-19
build: ensure image is built by default

kfundecmcsa:
We found after the repository went public that docker images are not
able to hosted publicly within the organization. This caused build
breaks since outside users were not able to pull the image and also
prevents CI from running in forks.
    
To restore functionality:
    
- add logic to ensure the image is built if it cannot be found locally
- in CI, add logic to see if the image cannot be pulled, indicating a
   fork, and if so, build the image accordingly
    
Additionally, the image repository name is duplicated in many places. To
reduce this duplication, have one source of truth for this value within
`docker/setupDockerEnv` to allow for sourcing by the various Docker
processes.
    
Refs: BARTON-250

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -61,9 +61,14 @@ jobs:
                     username: ${{ github.actor }}
                     password: ${{ secrets.GITHUB_TOKEN }}
             -
-                name: Pull barton_builder Docker image
+                name: Try to pull barton_builder Docker image
+                continue-on-error: true
                 run: |
-                     docker pull ghcr.io/rdkcentral/barton_builder:${{ needs.get-action-constants.outputs.image_tag }}
+                    if docker pull ${{ needs.get-action-constants.outputs.image_repo }}:${{ needs.get-action-constants.outputs.image_tag }}; then
+                        echo "Image pulled successfully"
+                    else
+                        echo "Could not pull image, will build locally via dockerw"
+                    fi
             -
                 name: Build BartonCore
                 run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/run-tests.yaml
```diff
@@ -61,9 +61,14 @@ jobs:
                     username: ${{ github.actor }}
                     password: ${{ secrets.GITHUB_TOKEN }}
             -
-                name: Pull barton_builder Docker image
+                name: Try to pull barton_builder Docker image
+                continue-on-error: true
                 run: |
-                     docker pull ghcr.io/rdkcentral/barton_builder:${{ needs.get-action-constants.outputs.image_tag }}
+                    if docker pull ${{ needs.get-action-constants.outputs.image_repo }}:${{ needs.get-action-constants.outputs.image_tag }}; then
```
tleacmcsa, line 67: This attempt to pull the image is in the "run-tests.yaml" which I suspect would come after the build finishes?  Sorta confused how the main build is getting the image from the repo.

kfundecmcsa: Not exactly. Each workflow that needs to do some stuff in the barton container will need to pull (or build) the image before doing anything in the container. This is a fundamental limitation to workflows. You can't, say, build the image in one workflow, then give it to another to do stuff with.

How we've set ours up to work (so far) is to pull/build as one of the first "steps", then all subsequent "steps" can use `dockerw` to enter that container and do a thing.

That being said, maybe `run-tests.yaml` wasn't the best name for this action. I tried to keep it fairly concise, rather than `build-barton-and-run-tests.yaml` or something, but maybe that would avoid confusion.

tleacmcsa: Yes, that is what confused me.  the "run-tests.yaml" file also builds.  I guess a rename would be nice.  Thanks.

cleithner-comcast (approved):

tleacmcsa (approved):

PR #19, opened 2025-05-23
build: Reorganize matter unit tests

cleithner-comcast:
build: Reorganize matter unit tests

Our matter unit tests were scattered about while other tests were
consistently in the core/test directory. This commit brings the
scattered tests under the core/test directory.


kfundecmcsa (approved):

core/test/src/subsystems/matter/CMakeLists.txt
```diff
@@ -28,10 +28,10 @@ include(BDSAddCppTest)
...
 
 file(GLOB TEST_SRC *.cpp)
-list(APPEND TEST_SRC ../ReadPrepareParamsBuilder.cpp
-                     ../MatterCommon.cpp
-                     ../providers/BartonCommissionableDataProvider.cpp
-                     ../providers/default/DefaultCommissionableDataProvider.cpp
+list(APPEND TEST_SRC ${CMAKE_SOURCE_DIR}/core/src/subsystems/matter/ReadPrepareParamsBuilder.cpp
+                     ${CMAKE_SOURCE_DIR}/core/src/subsystems/matter/MatterCommon.cpp
+                     ${CMAKE_SOURCE_DIR}/core/src/subsystems/matter/providers/BartonCommissionableDataProvider.cpp
+                     ${CMAKE_SOURCE_DIR}/core/src/subsystems/matter/providers/default/DefaultCommissionableDataProvider.cpp
```
rchowdcmcsa, line 34: CMAKE_SOURCE_DIR resolves to the directory where the top-level CMakeLists.txt resides, so it can be different when a client pulls Barton and tries to build. We should change these paths to something that'll be consistent across multiple projects.

core/test/src/subsystems/matter/CMakeLists.txt
```diff
@@ -43,10 +43,7 @@ pkg_check_modules(OPENSSL REQUIRED openssl)
 # Add more executables for truly isolated units that don't link to other libraries.
 bds_add_cpp_test(NAME testMatterSDKTooling
                 LIBS ${BDS_MATTER_LIB} ${OPENSSL_LINK_LIBRARIES} gmock crypto xhLog xhUtil xhDeviceDescriptors xhConfig
-                INCLUDES ${CMAKE_SOURCE_DIR}/core/src
-                         ${CMAKE_CURRENT_SOURCE_DIR}/..
-                         ${CMAKE_CURRENT_SOURCE_DIR}/../providers
-                         ${CMAKE_CURRENT_SOURCE_DIR}/../providers/default
+                INCLUDES ${BARTON_PRIVATE_INCLUDES}
                          ${CMAKE_BINARY_DIR}/matter-install/include/matter
                          ${CMAKE_SOURCE_DIR}/api/c/public
```
rchowdcmcsa, line 48: Should rewrite this to not use CMAKE_SOURCE_DIR.

rchowdcmcsa (changes requested):

core/test/src/subsystems/matter/CMakeLists.txt
cleithner-comcast: Done

cleithner-comcast:
> I dont agree with this change. Tests should be next to the code that it is testing.

I don't understand your statement. All of "core" has so far followed a mirror of src/* and test/* where things in test/*/foo are putting things in src/*/foo under test. A couple of matter tests were the only exception in that they were placed under src/ a la src/*/test/foo. Are you saying you would rather we go the opposite direction and take all the things under test/* and distribute them across src/* ?

cleithner-comcast:
> > > > > I dont agree with this change. Tests should be next to the code that it is testing.
> > > > 
> > > > 
> > > > I don't understand your statement. All of "core" has so far followed a mirror of src/* and test/* where things in test/_/foo are putting things in src/_/foo under test. A couple of matter tests were the only exception in that they were placed under src/ a la src/_/test/foo. Are you saying you would rather we go the opposite direction and take all the things under test/_ and distribute them across src/* ?
> > > 
> > > 
> > > ./core/test ./core/src/subsystems/matter/test ./core/deviceDrivers/matter/test ./api/c/test ./libs/time/c/test ./libs/json/c/test ./libs/util/c/test ./libs/url/c/test ./libs/config/c/test ./libs/types/c/test ./libs/concurrent/c/test ./libs/zhal/c/test ./libs/device/descriptors/c/test ./libs/device/philipsHue/c/test
> > > sure, lots of these are imported from their origin, but we should discuss this anyway.
> > 
> > 
> > In your list, every entry has a test/ directory for the entire target except the matter ones, which embed test directories among the source code directories. I'm only trying to retain consistency by pulling those tests out to the core/test dir.
> 
> I'm not aiming for consistency with something that we havent actually discussed or thought out. I'm starting with the question "how do we want this laid out" first. We may end up with all tests in a single directory, or we may decide that keeping tests next to the code that they test makes more sense. I'm in favor of the latter, but am open!

I see. I went down this route because I've personally found the discrepancy in organization worse to interact with than either of those two options, so I was trying to "right the plane" so to speak. Regarding preference, I've heavily interacted with projects of both and personally I dramatically prefer having tests in one place instead of scattered. Often, in a navigation window, I might try to navigate to some particular directory or file and having to scroll over a bunch of tests can be pretty annoying. Additionally, searching via ls, fdfind, or ag in a terminal for source only tokens is made extra annoying if there's scattered test/ directories as opposed to a singe one I can exclude easily enough. Automating testing for integration tests is easier when tests are in one place as well. This is primarily applies to "test discovery" frameworks like junit, pytest, or even gtest for unit testing.

tleacmcsa:
> > > > > > I dont agree with this change. Tests should be next to the code that it is testing.
> > > > > 
> > > > > 
> > > > > I don't understand your statement. All of "core" has so far followed a mirror of src/* and test/* where things in test/_/foo are putting things in src/_/foo under test. A couple of matter tests were the only exception in that they were placed under src/ a la src/_/test/foo. Are you saying you would rather we go the opposite direction and take all the things under test/_ and distribute them across src/* ?
> > > > 
> > > > 
> > > > ./core/test ./core/src/subsystems/matter/test ./core/deviceDrivers/matter/test ./api/c/test ./libs/time/c/test ./libs/json/c/test ./libs/util/c/test ./libs/url/c/test ./libs/config/c/test ./libs/types/c/test ./libs/concurrent/c/test ./libs/zhal/c/test ./libs/device/descriptors/c/test ./libs/device/philipsHue/c/test
> > > > sure, lots of these are imported from their origin, but we should discuss this anyway.
> > > 
> > > 
> > > In your list, every entry has a test/ directory for the entire target except the matter ones, which embed test directories among the source code directories. I'm only trying to retain consistency by pulling those tests out to the core/test dir.
> > 
> > 
> > I'm not aiming for consistency with something that we havent actually discussed or thought out. I'm starting with the question "how do we want this laid out" first. We may end up with all tests in a single directory, or we may decide that keeping tests next to the code that they test makes more sense. I'm in favor of the latter, but am open!
> 
> I see. I went down this route because I've personally found the discrepancy in organization worse to interact with than either of those two options, so I was trying to "right the plane" so to speak. Regarding preference, I've heavily interacted with projects of both and personally I dramatically prefer having tests in one place instead of scattered. Often, in a navigation window, I might try to navigate to some particular directory or file and having to scroll over a bunch of tests can be pretty annoying. Additionally, searching via ls, fdfind, or ag in a terminal for source only tokens is made extra annoying if there's scattered test/ directories as opposed to a singe one I can exclude easily enough. Automating testing for integration tests is easier when tests are in one place as well. This is primarily applies to "test discovery" frameworks like junit, pytest, or even gtest for unit testing.

Moving discussion to slack as this "thread" doesnt work well in github via quoting

tleacmcsa:
We discussed and I was convinced to move under a common test directory mostly around symbol use search filtering so symbol use in tests can be omitted when searching.

kfundecmcsa (approved):

rchowdcmcsa (dismissed):

rchowdcmcsa (approved):

tleacmcsa (approved):

PR #20, opened 2025-05-23
build: improve Docker tag management and container warnings

kfundecmcsa:
Add post-checkout hook to warn users when running in outdated containers after git operations like pull/checkout. This ensures developers are notified when their development environment needs to be rebuilt due to Docker image version requirements changing.

Also enhance Docker tag management to automatically use the highest versioned git tag (docker-builder-*) for image versioning, ensuring builds use compatible toolchains while handling custom tags gracefully. The system warns users when custom tags are based on outdated builders and provides guidance on updating, maintaining consistent environments across the team.

Refs: BARTON-246

### DIFF
--- a/docker/setupDockerEnv.sh
+++ b/docker/setupDockerEnv.sh
@@ -57,16 +57,55 @@ OUTFILE=$DIR/.env
 BARTON_TOP=$DIR/..
 IMAGE_REPO="ghcr.io/rdkcentral/barton_builder"
 
-# Find the highest versioned image tag reachable by HEAD that matches 'docker-builder-*'
-IMAGE_TAG=$(git tag -l 'docker-builder-*' --sort=-v:refname --merged | head -n 1 | sed 's/docker-builder-//')
+# Docker image version management
+#
+# This system automatically updates to the latest compatible Docker builder image
+# while handling user customizations in a predictable way:
+# - Always updates when a newer builder version is available to ensure builds
+#   use compatible toolchains and dependencies
+# - Warns when custom tags exist based on outdated builders
+# - Allows restoring custom tags after updates if needed
+#
+# The version tags are tied to git tags (docker-builder-*) to maintain consistency
+# between code and build environment versions.
 
-# Check if there is an image tag already defined in the .env file. If so, this could
-# imply the user has has defined a custom tag to use for the build process.
+# Find the highest versioned image tag reachable by HEAD that matches 'docker-builder-*'
+HIGHEST_BUILDER_TAG=$(git tag -l 'docker-builder-*' --sort=-v:refname --merged | head -n 1 | sed 's/docker-builder-//')
+IMAGE_TAG=$HIGHEST_BUILDER_TAG
+BUILDER_TAG_OUT_OF_DATE=false
+
 if [ -f "$OUTFILE" ]; then
-    if grep -q "IMAGE_TAG" "$OUTFILE"; then
-        IMAGE_TAG=$(grep "IMAGE_TAG" "$OUTFILE" | sed 's/IMAGE_TAG=//')
+
+    CURRENT_BUILDER_TAG=$(grep "CURRENT_BUILDER_TAG=" "$OUTFILE" | sed 's/CURRENT_BUILDER_TAG=//')
+
+    if [ "$HIGHEST_BUILDER_TAG" != "$CURRENT_BUILDER_TAG" ]; then
+        echo "Current docker-builder tag ($CURRENT_BUILDER_TAG) is out of date. The latest tag is $HIGHEST_BUILDER_TAG."
+        echo "The value of IMAGE_TAG in docker/.env will be updated to use latest version"
+        BUILDER_TAG_OUT_OF_DATE=true
     fi
+
+    IMAGE_TAG=$(grep "IMAGE_TAG" "$OUTFILE" | sed 's/IMAGE_TAG=//')
+
+    CUSTOM_TAG=false
+    if [ "$IMAGE_TAG" != "$CURRENT_BUILDER_TAG" ]; then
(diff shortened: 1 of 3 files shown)


hooks/post-checkout
kfundecmcsa, line 1: Missing a copyright header here. I'm suprised this got by our check and rdks? I'll add this in another commit

docker/setupDockerEnv.sh
```diff
@@ -57,17 +57,56 @@ OUTFILE=$DIR/.env
...
+    IMAGE_TAG=$(grep "IMAGE_TAG" "$OUTFILE" | sed 's/IMAGE_TAG=//')
+
+    CUSTOM_TAG=false
+    if [ "$IMAGE_TAG" != "$CURRENT_BUILDER_TAG" ]; then
+        CUSTOM_TAG=true
     fi
+
+    if [ "$CUSTOM_TAG" = true ] && [ "$BUILDER_TAG_OUT_OF_DATE" = true ]; then
+        echo "WARNING: The custom image tag '$IMAGE_TAG' is based on an older builder image ($CURRENT_BUILDER_TAG)."
+        echo "1. Rebase your docker related changes in '$IMAGE_TAG' to the latest version ($HIGHEST_BUILDER_TAG)"
```
cleithner-comcast, line 96: I think step 1 is incorrect. If we get here, we've already added a new docker version in the lineage due to either a rebase or a merge.

docker/setupDockerEnv.sh
```diff
@@ -80,8 +119,10 @@ echo "BUILDER_GID=$(id -g)" >> $OUTFILE
 # Save off the path to the Barton directory so we can mount it in the same path in the container
 echo "BARTON_TOP=$BARTON_TOP" >> $OUTFILE
 # Save off the image repo/tag into the .env file so it can be used in the compose process
```
cleithner-comcast, line 121: You didn't move this comment as well.

hooks/install.sh
```diff
@@ -29,7 +29,7 @@ MY_DIR=$(realpath $(dirname $0))
 TOP_DIR=${MY_DIR}/..
 HOOKS_DIR=${TOP_DIR}/.git/hooks
 
-HOOKS_SUPPORTED=(pre-commit)
+HOOKS_SUPPORTED=(pre-commit post-checkout)
```
cleithner-comcast, line 32: Good idea.

cleithner-comcast (changes requested):

docker/setupDockerEnv.sh
```diff
@@ -57,17 +57,56 @@ OUTFILE=$DIR/.env
...
-# Check if there is an image tag already defined in the .env file. If so, this could
-# imply the user has has defined a custom tag to use for the build process.
 if [ -f "$OUTFILE" ]; then
-    if grep -q "IMAGE_TAG" "$OUTFILE"; then
-        IMAGE_TAG=$(grep "IMAGE_TAG" "$OUTFILE" | sed 's/IMAGE_TAG=//')
+
+    CURRENT_BUILDER_TAG=$(grep "CURRENT_BUILDER_TAG=" "$OUTFILE" | sed 's/CURRENT_BUILDER_TAG=//')
+
+    if [ "$HIGHEST_BUILDER_TAG" != "$CURRENT_BUILDER_TAG" ]; then
+        echo "Current docker-builder tag ($CURRENT_BUILDER_TAG) is out of date. The latest tag is $HIGHEST_BUILDER_TAG."
```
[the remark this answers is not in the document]
cleithner-comcast: I interpret the logic to be very conservative. Any checkout in which lineage changes and builder versions have changed should force a rebuild/re-entry of the image version. That is, if main is 3.x and I go check out an old version of barton (for release maintenance), that checkout should force my builder version backwards in time. Kevin should verify that behavior empirically for confidence.

I am not sure what you are referring to with Jenkins, though.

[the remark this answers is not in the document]
kfundecmcsa: Keep in mind that the HIGHEST_BUILDER_TAG will only get the the highest `docker-builder-*` tag reachable from HEAD. In other words, it will only get the latest docker builder tag in the current lineage.

So to walk through the example you raised:

Let's say I'm on main, and the HIGHEST_BUILDER_TAG in this lineage is `docker-builder-3.0`

At this point, what's in docker/.env should be something like:

```
IMAGE_TAG=3.0
CURRENT_BUILDER_TAG=3.0
```

Later, we realized there is an issue with the 2.0 release, so we need to make a fix. We checkout the `release-2.0` branch, and in this branch, the HIGHEST_BUILDER_TAG in this lineage is `docker-builder-2.0`. When setupDockerEnv is ran, it will see that the HIGHEST_BUILDER_TAG != CURRENT_BUILDER_TAG (2.0 != 3.0) - then will always assume to use the HIGHEST_BUILDER_TAG.

So to answer this part of the question:

>If we release 2.0 and mainline moves to 3.0, then we have to go back to the 2.0 branch to make a 2.1 update, will this automatically update the 2.x branch we are on to the 3.x container which may not be compatible?

No, it will actually assume to use the older image - which is what we want.

But your point points out the flawed wording of some echo statements and variables. In this scenario you described, it shouldn't say things like "$CURRENT_BUILDER_TAG is out of date" and "updated to use latest version" - because we actually want to use older versions. I will make another commit to reword some of this stuff.

>Kevin should verify that behavior empirically for confidence.

Done. Works well

>We will certainly need to spin releases and builds off of older branches too. How will Jenkins handle that?

I assume you mean github actions here instead of Jenkins? If so, I'm not sure of all the details, but I would imagine this system would support it pretty well. The important part is to give the build system the SHA for the release, and it will find the highest docker-builder-* tag in the lineage and use that to make a build.

PR #30, opened 2025-06-10
build!: Refactor out public "device service" naming

cleithner-comcast:
This commit refactors the library targets of barton to BartonCore and BartonCoreStatic. Additionally, we don't attempt to install BartonCoreStatic as it doesn't make sense to install both the shared object and the static library. Clients relying on the static library should reference the cmake target directly or install it out of band.

In addition, this commit removes "device service" naming from public interfaces and cmake variables. The goal here is to standardize on the "barton core" nomenclature. Internal modules will still use "device service" if they already were.

Refs: BARTON-248


tleacmcsa (dismissed):
What a beast.  I spot checked randomly.

kfundecmcsa (approved):
I like this nomenclature change, nice!

rchowdcmcsa (dismissed):

tleacmcsa (approved):
Spot checked.

kfundecmcsa (approved):

mhughesacn:
Approved copyright as it is correct so no need to block but surprised to see a Copyright notice (or comment) in a json file

cleithner-comcast:
> Approved copyright as it is correct so no need to block but surprised to see a Copyright notice (or comment) in a json file

I am too. May have made sense in the initial private version for some reason but perhaps we could remove at some point. Thanks for checking.

PR #32, opened 2025-06-12
feat(matter): Client provided commissioning params

tleacmcsa:
This change unifies the discriminator and passcode used for commissioning between remote requests to open commissioning and internal/client invoked through the Barton API. In both cases the registered CommissionableDataProvider is used to provide these parameters.

This change also has the reference app set the discriminator and passcode to the same values as the ones used commonly in the Matter SDK if they are not already set.

Refs: BARTON-260

### DIFF
--- a/core/src/subsystems/matter/Matter.cpp
+++ b/core/src/subsystems/matter/Matter.cpp
@@ -1264,12 +1264,21 @@ bool Matter::OpenCommissioningWindow(chip::NodeId nodeId,
     // since we (or the remote device) are already commissioned, we can only support on network commissioning
     setupPayload.rendezvousInformation.SetValue(RendezvousInformationFlags(RendezvousInformationFlag::kOnNetwork));
 
-    uint16_t discriminator = Crypto::GetRandU16() % kMaxDiscriminator;
+    uint16_t discriminator = 0;
+    if (commissionableDataProvider->GetSetupDiscriminator(discriminator)!= CHIP_NO_ERROR)
+    {
+        icError("Failed to get setup discriminator from CommissionableDataProvider");
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


core/src/subsystems/matter/Matter.cpp
```diff
@@ -1264,12 +1264,21 @@ bool Matter::OpenCommissioningWindow(chip::NodeId nodeId,
     // since we (or the remote device) are already commissioned, we can only support on network commissioning
     setupPayload.rendezvousInformation.SetValue(RendezvousInformationFlags(RendezvousInformationFlag::kOnNetwork));
 
-    uint16_t discriminator = Crypto::GetRandU16() % kMaxDiscriminator;
+    uint16_t discriminator = 0;
+    if (commissionableDataProvider->GetSetupDiscriminator(discriminator)!= CHIP_NO_ERROR)
```
cleithner-comcast, line 1268: `)!=`

minor spacing :)

reference/src/device-service-reference-app.c
```diff
@@ -181,6 +182,34 @@ static gchar *getDefaultConfigDirectory(void)
     return confDir;
 }
 
+static void setDefaultParameters(BDeviceServiceInitializeParamsContainer *params)
```
cleithner-comcast, line 185: Are these really the first properties we've set in the reference app (besides allowed subsystems)?

cleithner-comcast (dismissed):

rchowdcmcsa (approved):

core/src/subsystems/matter/Matter.cpp
tleacmcsa: done

cleithner-comcast (approved):

rchowdcmcsa (approved):

PR #36, opened 2025-06-25
feat: make Matter BLE Controller device name configurable

kfundecmcsa:
Add BCORE_MATTER_BLE_CONTROLLER_DEVICE_NAME option to allow custom branding.

Refs: BARTON-253

### DIFF
--- a/config/cmake/options.cmake
+++ b/config/cmake/options.cmake
@@ -230,6 +230,11 @@ bcore_string_option(NAME BCORE_MATTER_DELEGATE_HEADER_PATHS
                   VALUE "${MATTER_PROVIDER_DELEGATE_PARENT_DIR}/delegates;
                          ${MATTER_DELEGATE_DEFAULT_DIR};")
 
+bcore_string_option(NAME BCORE_MATTER_BLE_CONTROLLER_DEVICE_NAME
+                    DEFINITION BARTON_CONFIG_MATTER_BLE_CONTROLLER_DEVICE_NAME
+                    DESCRIPTION "Name of the Matter BLE controller device."
+                    VALUE "BartonMatterBLEController")
+
 message(STATUS "- - - - - - - - - - - - - - - - ")
 
 message(STATUS "- - - - - - - - - - - - - - - - ")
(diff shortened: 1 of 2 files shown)


core/src/subsystems/matter/Matter.cpp
```diff
@@ -189,7 +189,6 @@ bool Matter::Init(uint64_t accountId, std::string &&attestationTrustStorePath)
         return false;
     }
     icDebug("Local node ID: 0x%" PRIx64, myNodeId);
-
```
tleacmcsa, line 192: I like whitespace, especially to break up unrelated chunks.  Any reason why you wanted to move the mkdir next to the unrelated node id log message?

config/cmake/options.cmake
```diff
@@ -230,6 +230,11 @@ bcore_string_option(NAME BCORE_MATTER_DELEGATE_HEADER_PATHS
                   VALUE "${MATTER_PROVIDER_DELEGATE_PARENT_DIR}/delegates;
                          ${MATTER_DELEGATE_DEFAULT_DIR};")
 
+bcore_string_option(NAME BCORE_MATTER_BLE_CONTROLLER_DEVICE_NAME
+                    DEFINITION BARTON_CONFIG_MATTER_BLE_CONTROLLER_DEVICE_NAME
+                    DESCRIPTION "Name of the Matter BLE controller device."
+                    VALUE "BartonMatterBLEController")
```
tleacmcsa, line 236: How about a default of "Matter-Controller" ?

core/src/subsystems/matter/Matter.cpp
kfundecmcsa: Nope - I added a debug line here to just print out the value of BLE_CONTROLLER_DEVICE_NAME to make sure what I added worked, then deleted the line before submission. Sorry about that - I'll add it back

cleithner-comcast (dismissed):

tleacmcsa (approved):

cleithner-comcast (approved):

PR #39, opened 2025-06-30
fix: prevent segfaults in subsystemManager

kfundecmcsa:
The MAP_FOREACH macro was causing segmentation faults when accessing
uninitialized stack variables in the subsystem manager.

To fix:
- Initialize all local variables in the macro and add a null check on the
map parameter to prevent crashes.
- Add null checks in all public functions that use MAP_FOREACH to enforce
proper API contracts and improve error handling.

### DIFF
--- a/core/src/subsystemManager.c
+++ b/core/src/subsystemManager.c
@@ -61,13 +61,17 @@ typedef struct
 #define MAP_FOREACH(value, map, expr)                                                                                  \
     do                                                                                                                 \
     {                                                                                                                  \
-        GHashTableIter iter;                                                                                           \
-        gpointer key, _value;                                                                                          \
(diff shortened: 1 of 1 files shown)


tleacmcsa (commented):
Instead of adding checks for subsystem == NULL everywhere, should we use the pthread_once_t initOnce that is defined in this module but for some reason not used?  There could be other things that should be initialized before public APIs are invoked as well.

tleacmcsa (approved):

kfundecmcsa:
Addressing these comments exposed some work to be done to improve the memory management in subsystemManager. This effort is out of scope for this PR and will be handled in a follow up PR.

cleithner-comcast (approved):

PR #41, opened 2025-07-08
fix(driver): correct values for illuminance and humidity

mkkoch:
Illuminance and humidity were being reported using the Zigbee encoded values rather than more user friendly units.

Refs: BARTON-268


tleacmcsa (approved):
Thanks!

kfundecmcsa (approved):

PR #48, opened 2025-07-18
build: Add BCoreMatterHelper

cleithner-comcast:
This commit provides a new cmake module to better interface to select and query for Matter providers and delegates. Currently, this implementations sits on top of the existing public options. This is because these options must exist still to allow clients a mechanism to select these providers/delegates via command-line arguments. As a result, the implementation of this module modifies these cache variables forcibly. For most places (including initial cache files), this is fine and you can freely use the module. For options.cmake, it is still required to define the defaults using bds_string_option for transparancy of defaults as well as for correct options dump in the cmake output. In addition, you would not want these defaults to force over client selected ones.

### DIFF
--- a/api/c/public/barton-core-properties.h
+++ b/api/c/public/barton-core-properties.h
@@ -191,6 +191,15 @@ G_BEGIN_DECLS
     B_CORE_BARTON_PREFIX B_CORE_MATTER_PREFIX "setupDiscriminator"
 
 /**
+ * MATTER_SETUP_PASSCODE: (value "barton.matter.setupPasscode")
+ *
+ * The setup passcode for the Matter device.
+ * Type: uint32
+ */
+#define B_CORE_BARTON_MATTER_SETUP_PASSCODE                                                                  \
+    B_CORE_BARTON_PREFIX B_CORE_MATTER_PREFIX "setupPasscode"
+
+/**
  * MATTER_SPAKE2P_ITERATION_COUNT: (value "barton.matter.spake2pIterationCount")
  *
  * The iteration count for the Matter device's spake2p verifier.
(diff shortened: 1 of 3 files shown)


tleacmcsa:
"This is because these options must exist still to allow clients a mechanism to select these providers/delegates via command-line arguments."  Is this true? Why must we preserve that capability, especially if it drives the way the module is built?

tleacmcsa (approved):
seems complicated, but I guess its needed... probably :-)

cleithner-comcast:
> seems complicated, but I guess its needed... probably :-)

I think it just seems complicated cause of the line count. It's really a lot of comments and pretty spacing and repetition (not exact, but approximate). I would say look at the usage in the linux.cmake file and see if that seems simple or complicated cause that's the contract I want to keep simple.

config/cmake/platforms/dev/linux.cmake
```diff
@@ -33,24 +33,15 @@ set(BCORE_BUILD_WITH_ASAN ON CACHE BOOL "Build with ASAN")
...
-    CACHE
-    STRING
-    "Provider header paths")
+
+bcore_matter_helper_delegate_select_implementation(
+    DEV SelfSignedCertifierOperationalCredentialsIssuer
+)
+
+bcore_matter_helper_provider_select_implementation(
+    DEFAULT DefaultCommissionableDataProvider
```
tleacmcsa, line 45: DEV and DEFAULT are not clearly documented/explained, or I missed it.

Are these functions about SELECTING an implementation, or adding options to be selected?  The function docs and name confuse me.

tleacmcsa: Not trying to be difficult, but it still doesnt explain WHAT a dev or default provider/implementation is or means.  All I can deduce is that these upper case names map to lower case directories in barton.

If you add an implementation do you also have to select it?

cleithner-comcast: I'm not really understanding what you are saying. Adding an implementation implies it is selected to me. I can understand the confusion in the operation signature because select might imply override, and so add is more descript. Both otherwise, they mean the same thing - you are choosing what matter plugin code source files are compiled into the library.

I'm not sure what the hang up is with dev/default because they make sense to me, but I think that's because I've created the structure. Source files under dev are for development. Otherwise we provide some default ones. If you want neither, provide your own (custom).

I'm also not trying to be difficult, I think I just am struggling to wrap my head around this as if it was new to me so if you have suggestions on how I can clear this up I'm all ears.

cleithner-comcast: You can specify as many as you want in each category, as described by the Usage in the documentation. It's completely flexible pick-and-choose what you want from within one call. A client should not select two providers that occupy the same functionality (example: two noc issuer delegates) but there's no safety nets within this cmake for that at this time.

tleacmcsa (approved):
I yield.

kfundecmcsa (changes requested):
Just a few typos to clean up to make the function descriptions a bit more clear

PR #49, opened 2025-07-24
feat: implement subsystem reference counting

kfundecmcsa:
Stacked PRs:
 * #50
 * __->__#49

--- --- ---

### feat: implement subsystem reference counting

Refactor Subsystem and SubsystemRegistration objects to use GLib's atomic
reference counting API. This makes subsystem memory management more intuitive
and safer through explicit ownership transfers.

Memory safety is improved by using g_atomic_rc_box functions and automatic
cleanup for deterministic object lifetime.

Refs: BARTON-267

### DIFF
--- a/core/src/subsystemManager.c
+++ b/core/src/subsystemManager.c
@@ -69,7 +69,8 @@ typedef struct
             g_hash_table_iter_init(&iter, (map));                                                                      \
             while (g_hash_table_iter_next(&iter, &key, &_value))                                                       \
             {                                                                                                          \
-                value = (SubsystemRegistration *) _value;                                                              \
+                g_autoptr(SubsystemRegistration) value =                                                               \
+                    subsystemRegistrationAcquire((SubsystemRegistration *) _value);                                    \
                 expr                                                                                                   \
             }                                                                                                          \
         }                                                                                                              \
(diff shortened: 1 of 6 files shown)


core/src/subsystemManager.h
```diff
@@ -123,6 +123,30 @@ typedef struct Subsystem
...
+ *
+ * @param subsystem pointer to reference counted Subsystem
+ */
+void releaseSubsystem(Subsystem *subsystem);
+
+/*
+ * Convenience macro to declare a scope bound Subsystem
+ */
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(Subsystem, releaseSubsystem)
+#define scoped_Subsystem g_autoptr(Subsystem)
```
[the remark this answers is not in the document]
kfundecmcsa: good call, I'll drop this

core/src/subsystemManager.c
```diff
@@ -83,10 +83,46 @@ static bool allDriversStarted = false;
...
+
 static void subsystemRegistrationDestroy(SubsystemRegistration *reg)
 {
-    pthread_mutex_destroy(&reg->mtx);
-    free(reg);
+    if (reg)
+    {
+        pthread_mutex_destroy(&reg->mtx);
+        releaseSubsystem((Subsystem *) reg->subsystem);
+        free(reg);
```
tleacmcsa, line 124: Doesnt this defeat the purpose of refcounting?  If this was the last reference, wouldnt this be a double free?  How would this code know for sure if nobody else had a reference to it and it was ok to free?  Pretty sure g_atomic_rc_box_release_full frees when no more references exist, but check.

core/src/subsystems/matter/matterSubsystem.cpp
```diff
@@ -121,17 +121,18 @@ static void accountIdChanged(const gchar *accountId);
...
-                                               .shutdown = matterSubsystemShutdown,
-                                               .onAllServicesAvailable = matterSubsystemAllServicesAvailable,
-                                               .onRestoreConfig = matterSubsystemRestoreConfig,
-                                               .onPostRestoreConfig = matterSubsystemPostRestoreConfig,
-                                               .getStatusJson = getStatusJson,
-                                               .name = MATTER_SUBSYSTEM_NAME,
-                                               .version = MATTER_SUBSYSTEM_VERSION};
-
-    subsystemManagerRegister(&matterSubsystem);
+    scoped_Subsystem matterSubsystem = createSubsystem();
```
tleacmcsa, line 124: when you switch this away from scoped_Subsystem, dont forget to release this pointer after registration or it will leak a refcount.

core/src/subsystemManager.c
cleithner-comcast: Yeah I rehydrated, not sure why I approved before. Here's the types at play:

1. The container (hashmap).
2. The members of the container (SubsystemRegistration)
3. The members of the SubsystemRegistration (Subsystem).

The container can be managed with a read/write lock and a NULL check. If a request comes in at the same time as subsystem manager is shutting down, the write lock in the shutdown will take precedence and the read lock acquisition by the request will get NULL back. The container lock must not be released while holding a pointer to the container, but that's fine since multiple requests can hold the read lock and this module is not particularly high traffic (no worry of starving write lock acquisition).

The members are a bit trickier. The SubsystemRegistration types contain their own mutexes, which provides synchronicity for their contents. However, there's not protection for themselves. If a request grabs a read lock on the container, pulls out a SubsystemRegistration reference, then releases the read lock, subsystemManagerShutdown could happen and pull the rug on that memory. Thus, the SubsystemRegistration type should be refcounted to prevent this.

The Subsystem type is interesting because it is created by subsystems and then registered to subsystemManager - the definition of a "shared type". A naive approach to this is have the subsystems give up that memory during registration. We generally do this for statically allocated memory, such as when we register drivers, but in those cases the type is lent to but not managed by the registry. A general rule of thumb is to minimize moving memory when you can - instead, acquire and release in the same scope (principle of scope-bound resource management). So if we wanted to be extra thorough, making Subsystem refcounted is necessary too. However, you can also establish the contract that anything registered is now owned by subsystemManager and never leak the Subsystem in your public contract anywhere else. This is _okay_ but you can't guarantee the client doesn't shoot themselves in the foot and keep a pointer to the Subsystem around. It's generally a bad contract because of that.

cleithner-comcast (changes requested):

core/src/subsystemManager.c
kfundecmcsa: >Doesnt this defeat the purpose of refcounting? If this was the last reference, wouldnt this be a double free?

uggh, yes. Accidentally left this in - thanks for catching it.

I'll get to work on getting the SubsystemRegistration objects refcounted as well. Thanks for the thoughtful comments!

tleacmcsa (dismissed):

cleithner-comcast (dismissed):
This looks good, but your branch is out of date. Since this is stack-pr based, you gotta do a rebase.

PR #50, opened 2025-07-24
feat: redesign subsystemManager with clear lifecycle

kfundecmcsa:
Stacked PRs:
 * __->__#50
 * #49

--- --- ---

### feat: redesign subsystemManager with clear lifecycle

Refactor subsystemManager to implement a consistent initialize/shutdown
lifecycle. Introduce a preregistration mechanism for subsystems to register
before manager initialization. This separates registration from
initialization and allows proper sequencing.

- Manage subsystem registration timing during initialization
- Improve subsystemManager tests with proper setup/teardown and
  auto-registered subsystem cleanup
- Add proper documentation for preregistration in subsystem interfaces

Refs: BARTON-267

### DIFF
--- a/core/src/subsystemManager.c
+++ b/core/src/subsystemManager.c
@@ -26,6 +26,7 @@
 
 #include "deviceService.h"
 #include "deviceServiceConfiguration.h"
+#include "glib.h"
 #include "icUtil/stringUtils.h"
 #include "provider/barton-core-property-provider.h"
 #define LOG_TAG     "deviceService"
@@ -76,14 +77,24 @@ typedef struct
         }                                                                                                              \
     } while (0)
 
-static pthread_once_t initOnce = PTHREAD_ONCE_INIT;
-
 static pthread_rwlock_t mutex = PTHREAD_RWLOCK_INITIALIZER;
 static GHashTable *subsystems;
 static bool allDriversStarted = false;
+static void subsystemPreRegistrationDestroy(void);
 static subsystemManagerReadyForDevicesFunc readyForDevicesCB = NULL;
 static void checkSubsystemForMigration(SubsystemRegistration *registration);
 static SubsystemRegistration *subsystemRegistrationAcquire(SubsystemRegistration *reg);
+static bool subsystemManagerInitialized = false;
+
+// Pre-registered subsystems are those that wish to be registered before the
+// subsystem manager is initialized. These are typically automatically registered
+// __attribute__((constructor)) functions in subsystem modules.
+static GList *preRegisteredSubsystems = NULL;
+
+__attribute__((constructor)) void registerPreRegistrationDestroy(void)
+{
+    atexit(subsystemPreRegistrationDestroy);
+}
 
 Subsystem *createSubsystem(void)
 {
@@ -253,17 +264,23 @@ static void onSubsystemDeinitialized(const char *subsystem)
 }
 
 /**
- * Called by atexit
+ * @brief Cleanup pre-registered subsystems at program exit.
  */
-static void unregisterSubsystems(void)
+static void subsystemPreRegistrationDestroy(void)
 {
-    g_hash_table_destroy(subsystems);
-    subsystems = NULL;
+    WRITE_LOCK_SCOPE(mutex);
+    g_list_free_full(preRegisteredSubsystems, (GDestroyNotify) releaseSubsystem);
+    preRegisteredSubsystems = NULL;
 }
 
-void subsystemManagerRegister(Subsystem *subsystem)
+/**
+ * Helper to register a subsystem to a specific hash table
+ *
+ * @param subsystem The subsystem to register
+ */
+static void subsystemManagerRegisterToTable(Subsystem *subsystem, GHashTable *table)
 {
-    if (subsystem == NULL || subsystem->initialize == NULL || subsystem->name == NULL)
+    if (subsystem == NULL || subsystem->initialize == NULL || subsystem->name == NULL || table == NULL)
     {
         icLogError(LOG_TAG, "%s: Can't register subsystem with unknown type or no initializer!", __func__);
         return;
@@ -271,26 +288,27 @@ void subsystemManagerRegister(Subsystem *subsystem)
 
     SubsystemRegistration *registration = createSubsystemRegistration(subsystem);
 
+    if (!g_hash_table_insert(table, (char *) subsystem->name, registration))
     {
-        WRITE_LOCK_SCOPE(mutex);
+        icLogWarn(LOG_TAG, "%s: unable to register subsystem '%s'!", __func__, subsystem->name);
+        subsystemRegistrationDestroy(registration);
+    }
+    else
+    {
+        icLogInfo(LOG_TAG, "subsystem '%s' registered", subsystem->name);
+    }
+}
 
-        if (subsystems == NULL)
-        {
-            subsystems =
-                g_hash_table_new_full(g_str_hash, g_str_equal, NULL, (GDestroyNotify) subsystemRegistrationMapDestroy);
-            atexit(unregisterSubsystems);
-        }
-
-        if (!g_hash_table_insert(subsystems, (char *) subsystem->name, registration))
-        {
-            subsystemRegistrationDestroy(registration);
-            registration = NULL;
-            icLogWarn(LOG_TAG, "%s: unable to register subsystem '%s'!", __func__, subsystem->name);
-        }
-        else
-        {
-            icLogInfo(LOG_TAG, "subsystem '%s' registered", subsystem->name);
-        }
+void subsystemManagerRegister(Subsystem *subsystem)
+{
+    WRITE_LOCK_SCOPE(mutex);
+    if (!subsystemManagerInitialized)
+    {
+        preRegisteredSubsystems = g_list_append(preRegisteredSubsystems, acquireSubsystem(subsystem));
+    }
+    else
+    {
+        subsystemManagerRegisterToTable(subsystem, subsystems);
     }
 }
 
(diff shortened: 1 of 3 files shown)


core/src/subsystemManager.h
```diff
@@ -210,8 +210,14 @@ void subsystemManagerEnterLPM(void);
 void subsystemManagerExitLPM(void);
 
 /**
- * Register your subsystem
- * @param subsystem A Subsystem that h
+ * Register your subsystem with the subsystem manager.
+ *
+ * If the subsystem manager hasn't been fully initialized yet, the subsystem will
```
tleacmcsa, line 215: This paragraph is an implementation detail that the caller does not need to know. You could just say that subsystems can be registered at any time.

tleacmcsa (approved):

core/src/subsystemManager.c
```diff
@@ -406,6 +424,29 @@ void subsystemManagerInitialize(subsystemManagerReadyForDevicesFunc readyForDevi
...
+    {
+        WRITE_LOCK_SCOPE(mutex);
+        if (!subsystemManagerInitialized)
+        {
+            // If subsystems exists but manager isn't initialized, we have an inconsistent state.
+            // Reset subsystems to ensure a clean initialization.
+            if (subsystems)
+            {
+                unregisterSubsystems();
+            }
```
cleithner-comcast, line 436: This is confusing. It would be like clearing a list member in a C++ object constructor with a similar comment. Trust that you have well-defined constructor/destructors.

core/src/subsystemManager.c
```diff
@@ -468,6 +510,9 @@ void subsystemManagerShutdown(void)
 
             if (registration->subsystem->shutdown != NULL) { registration->subsystem->shutdown(); });
     }
+
+    subsystemPreRegistrationDestroy();
```
cleithner-comcast, line 514: Having subsystemPreRegistrationDestroy in your destructor invalidates process lifecycle registration. The point of the preregistration list is to support process-lifecycle subsystems. I would simply never clear this list and, on initialize, add any members in here to the active subsystems list.

cleithner-comcast (changes requested):

core/src/subsystemManager.h
cleithner-comcast: I think I disagree. If you are going to have a "convenience" API that can do two different things based on state, it needs to be very clear to the caller what those things are. Maybe you don't need to be explicit that it's added to a list (because that is an implementation detail) but you should provide some assurance that this smart handles registering the subsystem regardless of subsystemManager's lifecycle state.

core/src/subsystemManager.c
```diff
@@ -75,13 +76,17 @@ typedef struct
...
 static GHashTable *subsystems;
 static bool allDriversStarted = false;
 static subsystemManagerReadyForDevicesFunc readyForDevicesCB = NULL;
 static void checkSubsystemForMigration(SubsystemRegistration *registration);
+static bool subsystemManagerInitialized = false;
+
+// Pre-registered subsystems are those that wish to be registered before the
+// subsystem manager is initialized. These are typically automatically registered
+// __attribute__((constructor)) functions in subsystem modules.
+static GList *preRegisteredSubsystems = NULL;
```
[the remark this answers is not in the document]
cleithner-comcast: Idk why github propagated this comment, I don't think it applies anymore.

core/src/subsystemManager.c
kfundecmcsa: I was trying to defend against a "shouldn't happen" scenario, which seems like good thing to do here. But I see what you mean that clearing the list is probably not the best way to do this. In fact, it could hide bugs in the logic. Maybe what would be better is to do `g_return_if_fail(subsystems == NULL);` here, what do you think?

core/src/subsystemManager.c
kfundecmcsa: I see what you mean. But the memory has to get freed eventually for the list. Should we maybe include an `atexit` call that finally frees the list? This would be kind of going against the original attempt to have the subsystemManager follow the init/shutdown lifecyle, but it could be only for the pre-registered list, so maybe it's not so bad.

core/src/subsystemManager.c
```diff
@@ -468,6 +510,9 @@ void subsystemManagerShutdown(void)
 
             if (registration->subsystem->shutdown != NULL) { registration->subsystem->shutdown(); });
     }
+
+    subsystemPreRegistrationDestroy();
+    unregisterSubsystems();
```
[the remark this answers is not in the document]
kfundecmcsa: We could update all the APIs to check the subsystemManagerInitialized a la `g_return_val_if_fail(subsystemManagerInitialized, false);` to resolve. What do you think?

core/src/subsystemManager.c
cleithner-comcast: I would rarely concern yourself with "shouldn't happen" scenarios from within the same module because you just write the module so "shouldn't happen" becomes "can't happen". If your constructor and destructor are atomic and indempotent, this instruction will never need to exist.

core/src/subsystemManager.c
```diff
@@ -253,44 +259,57 @@ static void onSubsystemDeinitialized(const char *subsystem)
...
+ * @brief Cleanup pre-registered subsystems at program exit.
  */
+static void subsystemPreRegistrationDestroy(void)
+{
+    WRITE_LOCK_SCOPE(mutex);
+    g_list_free_full(preRegisteredSubsystems, (GDestroyNotify) releaseSubsystem);
+    preRegisteredSubsystems = NULL;
+}
+
 static void unregisterSubsystems(void)
```
kfundecmcsa, line 271: This can be dropped now, will do before merge

core/src/subsystemManager.c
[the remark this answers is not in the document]
kfundecmcsa: I added logic to steal the globals, do the work locally, and move back to global in the init/shutdown functions. Please take a look and see what you think.

core/src/subsystemManager.c
```diff
@@ -410,32 +429,67 @@ void subsystemManagerInitialize(subsystemManagerReadyForDevicesFunc readyForDevi
...
+    SubsystemRegistration *registration = NULL;
+    MAP_FOREACH(registration, localSubsystems, checkSubsystemForMigration(registration););
+
+    // Move local work to globals atomically. We must do this BEFORE calling
+    // subsystem->initialize() because those functions will trigger callbacks
+    // that need to access the global subsystems table.
+    {
+        WRITE_LOCK_SCOPE(mutex);
+        readyForDevicesCB = readyForDevicesCallback;
+        subsystems = g_steal_pointer(&localSubsystems);
```
tleacmcsa, line 483: multiple interleaved/concurrent calls to subsystemManagerInitialize could leak here by blowing away a pointer, but it isnt very realistic.

tleacmcsa (approved):
tricky stuff.  hope its well tested!

core/src/subsystemManager.c
```diff
@@ -410,32 +429,67 @@ void subsystemManagerInitialize(subsystemManagerReadyForDevicesFunc readyForDevi
...
-            SubsystemRegistration *registration = NULL;
-            MAP_FOREACH(
-                registration, subsystems, if (strstr(disabledSubsystems, registration->subsystem->name) != NULL) {
-                    g_hash_table_iter_remove(&iter);
-                });
+            g_list_foreach(localPreRegisteredSubsystems, (GFunc) subsystemManagerRegisterToTable, localSubsystems);
         }
+
+        localSubsystemManagerInitialized = true;
+        atexit(subsystemPreRegistrationDestroy);
```
cleithner-comcast, line 454: Just create an attribute constructor function for registering this atexit call. Decouple process lifecycle from subsystem manager lifecycle.

core/src/subsystemManager.c
```diff
@@ -410,32 +429,67 @@ void subsystemManagerInitialize(subsystemManagerReadyForDevicesFunc readyForDevi
...
+    {
         WRITE_LOCK_SCOPE(mutex);
-        readyForDevicesCB = readyForDevicesCallback;
-        if (disabledSubsystems && subsystems)
+        localSubsystemManagerInitialized = subsystemManagerInitialized;
+        localPreRegisteredSubsystems = g_steal_pointer(&preRegisteredSubsystems);
+        localSubsystems = g_steal_pointer(&subsystems);
+    }
+
+    if (!localSubsystemManagerInitialized)
```
cleithner-comcast, line 443: subsystemManagerInitialize should be idempotent. If subsystemManagerInitialized flag is true, the function should do nothing at all. I would check this in the lock and just return.

core/src/subsystemManager.c
```diff
@@ -410,32 +429,67 @@ void subsystemManagerInitialize(subsystemManagerReadyForDevicesFunc readyForDevi
...
+    bool localSubsystemManagerInitialized = false;
 
+    {
         WRITE_LOCK_SCOPE(mutex);
-        readyForDevicesCB = readyForDevicesCallback;
-        if (disabledSubsystems && subsystems)
+        localSubsystemManagerInitialized = subsystemManagerInitialized;
+        localPreRegisteredSubsystems = g_steal_pointer(&preRegisteredSubsystems);
+        localSubsystems = g_steal_pointer(&subsystems);
+    }
```
cleithner-comcast, line 441: Keep in mind that you have just created a pseudo big lock by sleight of hand. By stealing the memory, you have removed the concurrency and made calls async failures (try again later) instead of blocking success. This isn't necessarily a bad thing depending on the interface (for example, a call to an API over the net), but in this case it's pretty whackadoo.

With that said, I know why you did this. It's kind of a shortcoming of the architecture in general. If we could start over with it, I'd just make subsystem manager a C++ class and the language behavior would cover a lot of issues.

cleithner-comcast (changes requested):

core/src/subsystemManager.c
kfundecmcsa: Done. This required a fair amount of restructuring of a couple tests, but I think it ultimately makes the tests more thorough now.

core/src/subsystemManager.c
kfundecmcsa: done, good call

core/src/subsystemManager.c
kfundecmcsa: thanks for the feedback, maybe one day we'll tackle that

PR #53, opened 2025-08-05
fix(thread): fix retrieving device role

tleacmcsa:
Coding error skipped translating and returning the OTBR device role to the caller.

Refs: BARTON-272


cleithner-comcast (approved):

kfundecmcsa (approved):
`fix` is in the title twice now - reads a little funny. Up to you if you wanna drop the second one

PR #61, opened 2025-08-20
fix: shutdown race in zhal

tleacmcsa:
There have been instances of a shutdown race in zhal due to a lock check gap between zhalTerm setting workerThreadRunning to false and the worker thread's while loop check of that flag. This change just collapses the shutdown checks inside the loop where there is no longer a gap.

Refs: BARTON-254


cleithner-comcast (commented):
Can the race condition be captured in a unit test? I suspect not.

PR #62, opened 2025-08-21
build: Set SO Version

cleithner-comcast:
This commit adds support for setting the library version and SO version of libBartonCore. It also defines a "barton API version" that is an analog for namespace/platform version used in gir generation.


tleacmcsa (approved):

kfundecmcsa (approved):
nice

PR #65, opened 2025-08-25
feat: implement software watchdog delegate interface

kfundecmcsa:
Add base BartonSoftwareWatchdogDelegate and Zigbee-specific extension to provide standardized watchdog functionality across subsystems.

- Create BartonSoftwareWatchdogDelegate with core watchdog operations (initialize, shutdown, tickle, getActionInProgress)
- Add ZigbeeSoftwareWatchdogDelegate extending base with Zigbee-specific recovery functions (updateWatchdogStatus, immediateRecoveryService)
- Implement thread-safe set-once pattern for delegate assignment in zigbeeSubsystem
- Add comprehensive documentation for both delegate interfaces

The delegate pattern allows subsystems to integrate with external watchdog systems while maintaining clean separation of concerns and enabling flexible watchdog implementations.

Resolves watchdog integration requirements for Zigbee subsystem.

### DIFF
--- a/api/c/public/private/softwareWatchdogDelegate/softwareWatchdogDelegate.h
+++ b/api/c/public/private/softwareWatchdogDelegate/softwareWatchdogDelegate.h
@@ -1,0 +1,125 @@
+// ------------------------------ tabstop = 4 ----------------------------------
+//
+// If not stated otherwise in this file or this component's LICENSE file the
+// following copyright and licenses apply:
+//
+// Copyright 2025 Comcast Cable Communications Management, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// ------------------------------ tabstop = 4 ----------------------------------
+
+//
+// Created by Kevin Funderburg on 8/22/2025.
+//
+
+#pragma once
+
+#include <stdbool.h>
+
+/**
+ * @brief Base software watchdog delegate interface for Barton Core subsystems.
+ *
+ * This struct provides a standardized interface for subsystems to interact with
+ * external software watchdog systems.
+ *
+ * The delegate pattern allows subsystems to:
+ * - Initialize and configure their watchdog monitoring
+ * - Periodically "tickle" the watchdog to indicate healthy operation
+ * - Check if recovery operations are in progress
+ * - Cleanly shutdown watchdog monitoring
+ *
+ * @note All function pointers may be NULL if the functionality is not needed
+ *       or not available. Always check for NULL before calling.
+ */
+typedef struct BartonSoftwareWatchdogDelegate
+{
+    /**
+     * @brief Initialize the software watchdog system.
+     *
+     * This function is called once during subsystem initialization to set up
+     * watchdog monitoring. It should configure any necessary watchdog entities,
+     * timers, or monitoring threads.
+     *
+     * @note May be NULL if no initialization is required
+     */
+    void (*initializeWatchdog)(void);
+
+    /**
+     * @brief Shutdown the software watchdog system.
+     *
+     * This function is called during subsystem shutdown to cleanly disable
+     * watchdog monitoring, stop any background threads, and free resources.
+     * Should ensure no false watchdog triggers occur during shutdown.
+     *
+     * @note May be NULL if no shutdown cleanup is required
+     */
+    void (*shutdownWatchdog)(void);
+
+    /**
+     * @brief Tickle/pet the watchdog to indicate healthy operation.
+     *
+     * @note May be NULL if not needed
+     */
+    void (*tickleWatchdog)(void);
+
+    /**
+     * @brief Check if a watchdog recovery action is currently in progress.
+     *
+     * This function allows subsystems to query whether the watchdog system
+     * is currently performing recovery operations (e.g., restarting services).
+     *
+     * @return true if recovery is in progress, false otherwise
+     * @note May be NULL if recovery status checking is not supported
+     */
+    bool (*getActionInProgress)(void);
+
+    /**
+     * @brief Name of the process being monitored by this watchdog.
+     *
+     * This string identifies the specific process or subsystem that this
+     * watchdog delegate is responsible for monitoring. Used for logging,
+     * debugging, and recovery coordination.
+     *
+     * @note REQUIRED - Must not be NULL
+     */
+    const char *processName;
+
+} BartonSoftwareWatchdogDelegate;
+
+/**
+ * @brief Create a new BartonSoftwareWatchdogDelegate instance.
+ *
+ * Allocates memory for a new watchdog delegate and initializes all function
+ * pointers to NULL. The caller is responsible for setting appropriate function
+ * pointers after creation.
+ *
+ * @param name The name of the process being monitored
+ * @return Pointer to newly allocated delegate, or NULL on error
+ * @note Caller must call destroyBartonSoftwareWatchdogDelegate() to free memory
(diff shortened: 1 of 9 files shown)


api/c/public/private/softwareWatchdogDelegate/softwareWatchdogDelegate.h
```diff
@@ -0,0 +1,125 @@
...
+ * The delegate pattern allows subsystems to:
+ * - Initialize and configure their watchdog monitoring
+ * - Periodically "tickle" the watchdog to indicate healthy operation
+ * - Check if recovery operations are in progress
+ * - Cleanly shutdown watchdog monitoring
+ *
+ * @note All function pointers may be NULL if the functionality is not needed
+ *       or not available. Always check for NULL before calling.
+ */
+typedef struct BartonSoftwareWatchdogDelegate
```
tleacmcsa, line 47: Please remove "Software" from this name and references like the docblock above.  That was only there originally to make it a bit more clear that it was different from Zilker's other watchdog (a service which is really systemd like).  Here it just makes the name too long and also doesnt add anything useful this it is no longer in the Zilker project.  A client may implement in hardware if they want... we dont force a software solution.

api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
```diff
@@ -384,6 +396,29 @@ uint64_t getLocalEui64(void);
  */
 int zigbeeSubsystemSetAddresses(void);
 
+/**
+ * Set the software watchdog delegate for zigbee subsystem operations.
+ *
+ * IMPORTANT: This must be called BEFORE zigbeeSubsystemInitialize().
+ * Once the subsystem is started, the delegate cannot be changed.
```
tleacmcsa, line 403:
```suggestion
 * Once the subsystem is initialized, the delegate cannot be set or changed.
```

api/c/public/private/subsystems/zigbee/zigbeeSubsystemWatchdogDelegate.h
```diff
@@ -0,0 +1,141 @@
...
+typedef enum
+{
+    ZIGBEE_CORE_RECOVERY_ENITITY_HEARTBEAT,
+    ZIGBEE_CORE_RECOVERY_ENITITY_COMM_FAIL,
+    ZIGBEE_CORE_RECOVERY_ENITITY_NETWORK_BUSY
+} ZigbeeCoreRecoveryEntity;
+
+extern const char *ZigbeeCoreRecoveryEntityLabels[];
+
+extern const char *zigbeeCoreRecoveryReasonLabels[];
```
tleacmcsa, line 44: the enum and labels should be defined next to each other, right?  Wherever that is (here or somewhere else).

core/src/softwareWatchdogDelegate/softwareWatchdogDelegate.c
```diff
@@ -0,0 +1,66 @@
...
+#define logFmt(fmt) "%s: " fmt, __func__
+
+BartonSoftwareWatchdogDelegate *createBartonSoftwareWatchdogDelegate(const char *name)
+{
+    BartonSoftwareWatchdogDelegate *retVal = NULL;
+
+    if (name)
+    {
+        retVal = calloc(1, sizeof(BartonSoftwareWatchdogDelegate));
+        if (retVal)
```
tleacmcsa, line 43: Please don't check to see if allocs return NULL.  It is impossible to handle correctly and only adds cruft.

core/src/softwareWatchdogDelegate/softwareWatchdogDelegate.c
```diff
@@ -0,0 +1,66 @@
...
+    {
+        icError("Process name is required");
+    }
+
+    return retVal;
+}
+
+void destroyBartonSoftwareWatchdogDelegate(BartonSoftwareWatchdogDelegate *delegate)
+{
+    if (delegate)
```
tleacmcsa, line 62: no need to check for NULL.  free(NULL) is fine.

api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
```diff
@@ -28,20 +28,32 @@
...
-#define NETWORK_BLOB_PROPERTY_NAME        "ZIGBEE_NETWORK_CONFIG_DATA"
+#define ZIGBEE_CORE_PROCESS_NAME               "ZigbeeCore"
+#define ZIGBEE_SUBSYSTEM_NAME                  "zigbee"
+#define NETWORK_BLOB_PROPERTY_NAME             "ZIGBEE_NETWORK_CONFIG_DATA"
+
+#define ZIGBEE_INCREMENT_COUNTERS_ON_NEXT_INIT "ZIGBEE_INCREMENT_COUNTERS_ON_NEXT_INIT"
+
+// The amount we should increment the counters after things like RMA.  The values here are what we have historically
+// used
+#define NONCE_COUNTER_INCREMENT_AMOUNT         0x1000
```
tleacmcsa, line 47: Why is stuff like this promoted from being private/hidden in the related .c file to part of an interface defined and exposed in a header file? I assume this would be part of Zilker's recovery logic and therefore doesn't have any purpose inside barton.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -771,80 +633,15 @@ static void zigbeeSubsystemSetUnready(void)
...
-    }
-
-    uint64_t localEui64 = loadLocalEui64();
-    scoped_generic char *region = b_core_property_provider_get_property_as_string(
-        propertyProvider, CPE_REGION_CODE_PROPERTY_NAME, "US");
+    bool retVal;
 
-    for (int i = 0; i < MAX_NETWORK_INIT_RETRIES; ++i)
-    {
-        int initResult = zhalNetworkInit(localEui64, region, blob, zhalProps);
```
tleacmcsa, line 827: This is some of the most critical and sensitive processing zigbeeSubsystem does : initialize the zigbee network.  It cannot be some optional piece of code to be owned by a client.

tleacmcsa (changes requested):
I was not able to go too far in it, but I left some comments as a start.

An important thing to think of is: a watchdog hook is optional.  The zigbee subsystem should work just fine without one at all.  It looks like some critical initialization code is being pulled out, which wont do.

I am not sure that the effort to make the watchdog a base class with an implementation for zigbee will pay off.  It is usually a good idea to hold off making stuff extensible like that until you have a second instance, at which point the refactoring takes place.  The odds of getting an abstract interface done correctly the first time with only a single implementation that will work for others is usually quite low and the effort can be expensive.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -613,55 +543,14 @@ static void zigbeeSubsystemShutdown(void)
...
-        if (networkBusySoftwareWatchdogCtx != NULL)
-        {
-            softwareWatchdogUnregister(networkBusySoftwareWatchdogCtx);
-            softwareWatchdogReleaseContext(networkBusySoftwareWatchdogCtx);
-            networkBusySoftwareWatchdogCtx = NULL;
-        }
-#endif // BARTON_CONFIG_SUPPORT_SOFTWARE_WATCHDOG
-    }
+    destroyZigbeeSoftwareWatchdogDelegate(zigbeeSoftwareWatchdogDelegate);
+    zigbeeSoftwareWatchdogDelegate = NULL;
```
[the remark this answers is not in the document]
tleacmcsa: If you put a mutex here, where would the other mutex use be?  In other words: a mutex at only one point doesn't do anything.

core/src/subsystems/zigbee/zigbeeSubsystem.c
tleacmcsa: `// Start our repeating task which, when successful, pets the software watchdog for ZigbeeCore and
 // ensures the network is initialized.  We will do this even if the software watchdog configuration fails
 // since it takes care of network health as well`

This comment block is the start of the problem.  It describes this thing (the task) that is responsible for two different things which is kinda sucky (and my fault).

Checkout the ZIGBEE_WATCHDOG_ENABLED_PROP check at the top of configureWatchdogs.  If this was false, we wouldnt initialize the network at all and zigbee would be broken.  Many problems here that were before your changes.  This is getting a bit risky for this release.

kfundecmcsa:
I made yet another PR for this one based on a branch Thomas made. Closing this one.

PR #67, opened 2025-08-27
feat(zigbee): implement watchdog delegate interface

kfundecmcsa:
Add ZigbeeWatchdogDelegate interface to externalize watchdog operations for the Zigbee subsystem. This allows clients to provide custom watchdog implementations for monitoring ZigbeeCore health and recovery actions.

Features:
- ZigbeeWatchdogDelegate struct
- Delegate registration with comprehensive validation
- ZHAL response event forwarding to delegate
- Watchdog entity monitoring (heartbeat, comm fail, network busy)
- Recovery action coordination (restart, pet watchdog, status updates)
- Device communication restoration notifications

The delegate handles all watchdog logic including state tracking and recovery decisions. The Zigbee subsystem now focuses solely on core functionality while delegating health monitoring to external implementations.

### DIFF
--- a/api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
+++ b/api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
@@ -28,6 +28,7 @@
 #define FLEXCORE_ZIGBEESUBSYSTEM_H
 
 #include "device-driver/device-driver.h"
+#include "private/subsystems/zigbee/zigbeeWatchdogDelegate.h"
 #include "zigbeeAttributeTypes.h"
 #include <cjson/cJSON.h>
 #include <deviceDescriptor.h>
@@ -35,6 +36,7 @@
 #include <stdbool.h>
 #include <zhal/zhal.h>
 
+#define ZIGBEE_CORE_PROCESS_NAME          "ZigbeeCore"
 #define ZIGBEE_SUBSYSTEM_NAME             "zigbee"
 #define NETWORK_BLOB_PROPERTY_NAME        "ZIGBEE_NETWORK_CONFIG_DATA"
 
@@ -161,6 +163,20 @@ int zigbeeSubsystemUnregisterDiscoveryHandler(ZigbeeSubsystemDeviceDiscoveredHan
 int zigbeeSubsystemRegisterDeviceListener(uint64_t eui64, ZigbeeSubsystemDeviceCallbacks *callbacks);
 
 int zigbeeSubsystemUnregisterDeviceListener(uint64_t eui64);
+
+/**
+ * Set the watchdog delegate for zigbee subsystem operations.
+ *
+ * All function pointers in the delegate structure are mandatory and will be
+ * validated during registration. If any function pointer is NULL, the delegate
+ * will be rejected.
+ *
+ * IMPORTANT: This must be called BEFORE zigbeeSubsystemInitialize().
+ * Once the subsystem is initialized, the delegate cannot be set or changed.
+ *
+ * @param delegate The watchdog delegate implementation
+ */
+void zigbeeSubsystemSetWatchdogDelegate(ZigbeeWatchdogDelegate *delegate);
 
 // increment the discovering device count.  Discovery runs until the count is decremented to zero
 int zigbeeSubsystemStartDiscoveringDevices(void);
--- a/api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
+++ b/api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
@@ -1,0 +1,180 @@
+// ------------------------------ tabstop = 4 ----------------------------------
+//
+// If not stated otherwise in this file or this component's LICENSE file the
+// following copyright and licenses apply:
+//
+// Copyright 2025 Comcast Cable Communications Management, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// ------------------------------ tabstop = 4 ----------------------------------
+
+//
+// Created by Kevin Funderburg on 8/26/2025.
+//
+
+#pragma once
+
+#include <stdbool.h>
+#include <zhal/zhal.h>
+
+typedef enum
+{
+    ZIGBEE_CORE_RECOVERY_ENITITY_HEARTBEAT,
+    ZIGBEE_CORE_RECOVERY_ENITITY_COMM_FAIL,
+    ZIGBEE_CORE_RECOVERY_ENITITY_NETWORK_BUSY
+} ZigbeeCoreRecoveryEntity;
+
+static const char *ZigbeeCoreRecoveryEntityLabels[] = {"heartbeat", "communication failure", "network busy"};
+
+static const char *zigbeeCoreRecoveryReasonLabels[] = {"Recovery reason: heartbeat",
+                                                       "Recovery reason: communication failure",
+                                                       "Recovery reason: network busy"};
+
+typedef enum
+{
+    ZIGBEE_CORE_RESTART_ACTIVE,     // Delegate is attempting restart (keep trying)
+    ZIGBEE_CORE_RESTART_NOT_ACTIVE, // Delegate won't/can't restart (stop trying)
+} ZigbeeCoreRestartStatus;
+
+/**
+ * @brief Watchdog delegate interface for Zigbee subsystem monitoring.
+ *
+ * This structure defines the interface that external watchdog implementations
+ * must provide to monitor and manage the health of the Zigbee subsystem and
+ * ZigbeeCore process. The Zigbee subsystem uses this delegate to report
+ * health status, request recovery actions, and coordinate with system-wide
+ * watchdog functionality.
+ *
+ * @note ALL function pointers in this structure are MANDATORY and must be
+ *       implemented. The Zigbee subsystem will reject any delegate that has
+ *       NULL function pointers during registration.
+ *
+ * @see zigbeeSubsystemSetWatchdogDelegate()
+ */
+typedef struct ZigbeeWatchdogDelegate
+{
+    /**
+     * @brief Initialize the software watchdog system.
+     *
+     * This function is called once during subsystem initialization to set up
+     * watchdog monitoring. It should configure any necessary watchdog entities,
+     * timers, or monitoring threads.
+     *
+     */
+    void (*initializeWatchdog)(void);
+
+    /**
+     * @brief Shutdown the software watchdog system.
+     *
+     * This function is called during subsystem shutdown to cleanly disable
+     * watchdog monitoring and free resources.
+     */
+    void (*shutdownWatchdog)(void);
+
+    /**
+     * @brief Update the status of a specific Zigbee watchdog entity.
+     *
+     * This function allows the Zigbee subsystem to report the health status of
+     * various monitored entities (heartbeat, network busy, communication failure).
+     *
+     * @param ZigbeeCoreRecoveryEntity The specific Zigbee entity being reported on
+     * @param status true if the entity is healthy, false if it's in a failure state
+     * @return true if the status update was processed successfully, false on error
+     */
+    bool (*updateWatchdogStatus)(ZigbeeCoreRecoveryEntity watchdogEntity, bool status);
+
+    /**
+     * @brief Pet the watchdog for a specific entity.
+     *
+     * This function is called to indicate that a specific Zigbee entity is operating
+     * normally and should reset its watchdog timer. This prevents the watchdog from
+     * triggering recovery actions for healthy entities.
+     *
+     * @param watchdogEntity The specific entity to pet
+     * @return true if the pet was successful, false on error
+     */
+    bool (*petWatchdog)(ZigbeeCoreRecoveryEntity watchdogEntity);
+
+    /**
+     * @brief Restart the ZigbeeCore process.
+     *
+     * This function is called when the Zigbee subsystem determines that ZigbeeCore
+     * needs to be restarted due to failure conditions. The implementation should
+     * attempt to restart the ZigbeeCore process and return the appropriate status.
+     *
+     * @return ZIGBEE_CORE_RESTART_ACTIVE if restart is being attempted,
+     *         ZIGBEE_CORE_RESTART_NOT_ACTIVE if restart cannot or will not be performed
+     */
+    ZigbeeCoreRestartStatus (*restartZigbeeCore)(void);
+
+    /**
+     * @brief Notify that device communication has been restored.
+     *
+     * This function is called when devices that were previously in communication
+     * failure have successfully resumed communication. This allows the watchdog
+     * system to update its state and potentially cancel pending recovery actions.
+     *
+     * @return true if the notification was processed successfully, false on error
+     */
+    bool (*notifyDeviceCommRestored)(void);
+
+    /**
+     * @brief Handle ZHAL response events for watchdog monitoring.
+     *
+     * This function is called for each ZHAL operation response to monitor
+     * the health of the Zigbee stack.
+     *
+     * @param responseType The type of ZHAL operation that completed
+     * @param resultCode The ZHAL_STATUS result code from the operation
+     */
+    void (*zhalResponseHandler)(const char *responseType, ZHAL_STATUS resultCode);
+
+    /**
+     * @brief Check if a watchdog recovery action is currently in progress.
+     *
+     * This function allows subsystems to query whether the watchdog system
+     * is currently performing recovery operations (e.g., restarting services).
+     *
+     * @return true if recovery is in progress, false otherwise
+     */
+    bool (*getActionInProgress)(void);
+
+} ZigbeeWatchdogDelegate;
+
+/**
+ * @brief Create a new ZigbeeWatchdogDelegate instance.
+ *
+ * Allocates memory for a new watchdog delegate and initializes all function
+ * pointers to NULL. The caller is responsible for setting appropriate function
+ * pointers after creation.
+ *
+ * @param name The name of the process being monitored
+ * @return Pointer to newly allocated delegate
+ * @note Caller must call destroyZigbeeWatchdogDelegate() to free memory
+ */
+ZigbeeWatchdogDelegate *createZigbeeWatchdogDelegate(void);
+
+/**
+ * @brief Destroy a ZigbeeWatchdogDelegate instance.
+ *
+ * Frees the memory allocated for the watchdog delegate. Does not call any
+ * of the function pointers (like shutdownWatchdog) - the caller should
+ * perform any necessary cleanup before calling this function.
+ *
+ * @param delegate Pointer to the delegate to destroy, may be NULL (no-op)
+ * @note After calling this function, the delegate pointer is invalid
+ */
+void destroyZigbeeWatchdogDelegate(ZigbeeWatchdogDelegate *delegate);
--- a/config/cmake/options.cmake
+++ b/config/cmake/options.cmake
@@ -182,13 +182,6 @@ bcore_option(NAME BCORE_M1LTE
 bcore_option(NAME BCORE_SETUP_WIZARD
            DEFINITION BARTON_CONFIG_SETUP_WIZARD
            DESCRIPTION "Support for behavioral changes if there is an \"activation period\"")
-bcore_option(NAME BCORE_SUPPORT_SOFTWARE_WATCHDOG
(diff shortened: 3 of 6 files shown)


core/src/subsystems/zigbee/zigbeeWatchdogDelegate.c
```diff
@@ -0,0 +1,49 @@
...
+ZigbeeWatchdogDelegate *createZigbeeWatchdogDelegate(void)
+{
+    ZigbeeWatchdogDelegate *retVal = calloc(1, sizeof(ZigbeeWatchdogDelegate));
+    retVal->initializeWatchdog = NULL;
+    retVal->shutdownWatchdog = NULL;
+    retVal->updateWatchdogStatus = NULL;
+    retVal->petWatchdog = NULL;
+    retVal->restartZigbeeCore = NULL;
+    retVal->notifyDeviceCommRestored = NULL;
+    retVal->getActionInProgress = NULL;
```
kfundecmcsa, line 41: forgot `zhalResponseHandler` here

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+{
+    ZIGBEE_CORE_RESTART_ACTIVE,     // Delegate is attempting restart (keep trying)
+    ZIGBEE_CORE_RESTART_NOT_ACTIVE, // Delegate won't/can't restart (stop trying)
+} ZigbeeCoreRestartStatus;
+
+/**
+ * @brief Watchdog delegate interface for Zigbee subsystem monitoring.
+ *
+ * This structure defines the interface that external watchdog implementations
+ * must provide to monitor and manage the health of the Zigbee subsystem and
```
tleacmcsa, line 56:
```suggestion
 * may provide to monitor and manage the health of the Zigbee subsystem.
```

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+ * ZigbeeCore process. The Zigbee subsystem uses this delegate to report
```
tleacmcsa, line 57:
```suggestion
 * The Zigbee subsystem uses this delegate to report
```

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+ * health status, request recovery actions, and coordinate with system-wide
```
tleacmcsa, line 58: What is meant by "system-wide watchdog functionality" ?

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+ * @note ALL function pointers in this structure are MANDATORY and must be
+ *       implemented. The Zigbee subsystem will reject any delegate that has
+ *       NULL function pointers during registration.
+ *
+ * @see zigbeeSubsystemSetWatchdogDelegate()
+ */
+typedef struct ZigbeeWatchdogDelegate
+{
+    /**
+     * @brief Initialize the software watchdog system.
```
tleacmcsa, line 70: There should be no mention of "software" watchdog.  Moreover, this only initializes the delegate.  If the delegate's implementation does more grand things like initializing some supporting system, thats a private detail of its implementation.

tleacmcsa (changes requested):

api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
```diff
@@ -162,6 +164,20 @@ int zigbeeSubsystemRegisterDeviceListener(uint64_t eui64, ZigbeeSubsystemDeviceC
 
 int zigbeeSubsystemUnregisterDeviceListener(uint64_t eui64);
 
+/**
+ * Set the watchdog delegate for zigbee subsystem operations.
```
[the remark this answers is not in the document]
kfundecmcsa: Great point, done.

kfundecmcsa (commented):
I addressed most of the comments, there is some open questions on a couple.

api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
```diff
@@ -28,13 +28,15 @@
...
 #include "device-driver/device-driver.h"
+#include "private/subsystems/zigbee/zigbeeWatchdogDelegate.h"
 #include "zigbeeAttributeTypes.h"
 #include <cjson/cJSON.h>
 #include <deviceDescriptor.h>
 #include <icTypes/icLinkedList.h>
 #include <stdbool.h>
 #include <zhal/zhal.h>
 
+#define ZIGBEE_CORE_PROCESS_NAME          "ZigbeeCore"
```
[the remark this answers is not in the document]
kfundecmcsa: Good call, done.

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+//
+// Created by Kevin Funderburg on 8/26/2025.
+//
+
+#pragma once
+
+#include <stdbool.h>
+#include <zhal/zhal.h>
+
+typedef enum
```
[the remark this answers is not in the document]
kfundecmcsa: At the time when I added these, both barton and zilker had dependencies on these, which is why they were here. But now only zilker does, so it makes total sense to remove them.

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,137 @@
...
+ */
+typedef struct ZigbeeWatchdogDelegate
+{
+    /**
+     * @brief Shutdown the delegate.
+     *
+     * This function is called during subsystem shutdown to cleanly disable
+     * watchdog monitoring and free resources.
+     */
+    void (*shutdown)(void);
```
tleacmcsa, line 62: dont shoot me, but maybe init would be useful just for symmetry. :-) .

Earlier it is stated that all functions must be set.  It is probably better to remove that comment and instead indicate which function pointers are optional and which are mandatory.  init and shutdown, for example, shouldnt be mandatory.

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @brief Immediately update the status of the 'network busy' watchdog entity.
```
tleacmcsa, line 65: "Immediately" is left over from when there were multiple ways to do it (pet vs. set state).  It could be dropped now I think.  Nit.

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param isHealthy true if the entity is healthy, false if it's in a failure state
+     * @return true if the status update was processed successfully, false on error
+     */
+    bool (*updateNetworkBusyStatus)(bool isHealthy);
```
tleacmcsa, line 70: could clarify the arg name (again, dont shoot me!).  since we now have distinct and specific calls, it could be isBusy.

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @brief Immediately update the status of the 'comm fail' watchdog entity.
+     *
+     * @param isHealthy true if the entity is healthy, false if it's in a failure state
+     * @return true if the status update was processed successfully, false on error
+     */
+    bool (*updateCommFailStatus)(bool isHealthy);
```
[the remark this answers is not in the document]
kfundecmcsa: done

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
kfundecmcsa: done

kfundecmcsa (commented):
I responded to several of these comments saying "done", but I haven't pushed the code yet. I have some open questions I'd like to see how they play out before I push another commit as it may change some stuff.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -4137,7 +4186,32 @@ static void *checkAllDevicesInCommThreadProc(void *arg)
...
+
+            if (watchdogDelegate)
+            {
+                recoveryInProgress = watchdogDelegate->getActionInProgress();
+            }
+
+            // update status only when no recovery is in progress i.e. zigbeeCore is not in restarting phase
+            if (!recoveryInProgress)
+            {
+                watchdogDelegate->updateCommFailStatus(false);
```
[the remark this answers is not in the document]
kfundecmcsa: done

api/c/public/private/subsystems/zigbee/zigbeeWatchdogDelegate.h
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * @brief Handle ZHAL response events for watchdog monitoring.
+     *
+     * This function is called for each ZHAL operation response to monitor
+     * the health of the Zigbee stack.
+     *
+     * @param responseType The type of ZHAL operation that completed
+     * @param resultCode The ZHAL_STATUS result code from the operation
+     */
+    void (*zhalResponseHandler)(const char *responseType, ZHAL_STATUS resultCode);
```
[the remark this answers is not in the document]
tleacmcsa: networkGoodLast is just a local static variable to keep track of the previous network busy state.  This could just move to the delegate since its the one that would get told about the network busy condition.  It also already knows if recovery is in progress.  The responseHandler would just always tell the delegate the isNetworkBusy value each time responseHandler is called with either of the two tracked responseTypes.  the delegate has access to the recovery in progress and then can track the networkGoodLast itself.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -441,13 +427,28 @@ static void waitForInitialZigbeeCoreStartup(void)
         // try one more time
         if (zhalHeartbeatRc != 0)
         {
-            //WATCHDOG: let the delegate know that it should restart ZigbeeCore.  If it does
-            //  it should return a code we can use to know that we should keep trying in this
-            //  loop, otherwise if there is no watchdog delegate or it says it isnt doing anything
-            //  about it, we break
-            // if (delegate wants to restart ZigbeeCore) zigbeeCoreRestartCount++;
+            bool keepTryingRestart = true;
```
tleacmcsa, line 430: local variable not needed.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -457,6 +458,40 @@ static void waitForInitialZigbeeCoreStartup(void)
     }
 }
 
+void zigbeeSubsystemSetWatchdogDelegate(ZigbeeWatchdogDelegate *delegate)
```
tleacmcsa, line 461: The caller has no way to know if the delegate was valid/used.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -457,6 +458,40 @@ static void waitForInitialZigbeeCoreStartup(void)
...
+    bool validDelegate = false;
+
+    if (delegate)
+    {
+        // init and shutdown are optional, all others are mandatory
+        validDelegate = delegate->setAllDevicesInCommFail && delegate->petZhal && delegate->restartZhal &&
+                        delegate->zhalResponseHandler && delegate->getActionInProgress;
+    }
+
+    mutexLock(&watchdogDelegateMtx);
```
tleacmcsa, line 473: Let's be clear on the locking strategy.  As-is, this mutex only protects against two threads trying to set the delegate at the same time.  Given ownership transfer and the requirement for the delegate to be set before initialization, where are the potential races and the data that we need to protect?  Maybe here and the shutdown?

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -4137,7 +4195,33 @@ static void *checkAllDevicesInCommThreadProc(void *arg)
 
         if (devicesInCommFail)
         {
-            //WATCHDOG: let the delegate know that all zigbee devices are in comm fail
+            bool recoveryInProgress = false;
+
+            if (watchdogDelegate)
+            {
+                // The order of the operations here is important. We need to first check if a recovery is in progress
```
tleacmcsa, line 4202: This comment doesnt make sense.  It implies that reading the action in progress state somehow has a side effect that makes the setAllDevicesInCommFail function do something different, which must not be the case.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -4137,7 +4195,33 @@ static void *checkAllDevicesInCommThreadProc(void *arg)
...
+                // before we set the "all devices in comm fail" status as this will kick off recovery actions.
+                recoveryInProgress = watchdogDelegate->getActionInProgress();
+                watchdogDelegate->setAllDevicesInCommFail(true);
+            }
+
+            // update status only when no recovery is in progress i.e. zigbeeCore is not in restarting phase
```
tleacmcsa, line 4208: stale comment.  All we are doing is adjusting the fast comm fail stuff

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -4546,20 +4630,23 @@ void zigbeeSubsystemDeviceAnnounced(uint64_t eui64, zhalDeviceType deviceType, z
 
 static void responseHandler(const char *responseType, ZHAL_STATUS resultCode)
 {
-    if (responseType == NULL)
```
tleacmcsa, line 4549: Hmm, since responseType is a pointer, not a bool, I'm not sure how I feel about this change.  Of course it works, but I don't think this is our style and it isnt technically correct.

tleacmcsa (commented):
Really close!

kfundecmcsa:
It occurred to me to add a unit test to ensure `zigbeeSubsystemSetWatchdogDelegate` works as expected, please let me know what you think.

tleacmcsa (approved):
FYI, I was on the fence all along about not making any of the delegate callbacks required.   We dont need to enforce it.  If they dont provide a petZhal function, then we dont call it.  no biggie.. its up to them.  That would make the unit tests and some code more simple.  I would still generally prefer that, but its up to you and I'm ok with it as it is.

api/c/public/private/subsystems/zigbee/zigbeeSubsystem.h
```diff
@@ -162,6 +163,25 @@ int zigbeeSubsystemRegisterDeviceListener(uint64_t eui64, ZigbeeSubsystemDeviceC
...
+ * provides monitoring capabilities for network communication status, busy states,
+ * and automatic recovery mechanisms for stack failures.
+ *
+ * IMPORTANT:
+ * - This must be called BEFORE zigbeeSubsystemInitialize().
+ * - Once the subsystem is initialized, the delegate cannot be set or changed.
+ * - A delegate that has NULL mandatory function pointers will be rejected
+ *   during registration.
+ *
+ * @param delegate The watchdog delegate implementation (ownership transferred)
```
cleithner-comcast, line 180: No action requested, but avoid ownership transfer contracts. I know this is meant to fit to private needs and not become a public interface so I'm not asking for anything different here.

config/cmake/options.cmake
```diff
@@ -182,13 +182,6 @@ bcore_option(NAME BCORE_M1LTE
 bcore_option(NAME BCORE_SETUP_WIZARD
            DEFINITION BARTON_CONFIG_SETUP_WIZARD
            DESCRIPTION "Support for behavioral changes if there is an \"activation period\"")
-bcore_option(NAME BCORE_SUPPORT_SOFTWARE_WATCHDOG
-           DEFINITION BARTON_CONFIG_SUPPORT_SOFTWARE_WATCHDOG
-           DESCRIPTION "Support for zigbee watchdog feature using the software watchdog library.")
-bcore_int_option(NAME BCORE_SOFTWARE_WATCHDOG_TROUBLE_CODE_ZIGBEE_CORE
-              DEFINITION BARTON_CONFIG_SOFTWARE_TROUBLE_CODE_ZIGBEE_CORE_WATCHDOG
-              DESCRIPTION "The trouble code to use when reporting a zigbee core failure to the software watchdog library."
-              VALUE 11)
```
cleithner-comcast, line 191: nice!

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -441,13 +427,24 @@ static void waitForInitialZigbeeCoreStartup(void)
...
-            //  loop, otherwise if there is no watchdog delegate or it says it isnt doing anything
-            //  about it, we break
-            // if (delegate wants to restart ZigbeeCore) zigbeeCoreRestartCount++;
-
-            // NOTE: BARTON_CONFIG_SOFTWARE_TROUBLE_CODE_ZIGBEE_CORE_WATCHDOG should not be needed in barton anymore
+            if (watchdogDelegate)
+            {
+                if (watchdogDelegate->restartZhal() == ZHAL_RESTART_ACTIVE)
+                {
+                    zigbeeCoreRestartCount++;
```
cleithner-comcast, line 434: It seems like zigbeeSusbystem should not have references to "ZigbeeCore" in here anymore, right? At least not in the context of the watchdog functionality.

Similar comment for the logging but be mindful of logs that may have been tracked by telemetry before.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -457,6 +454,49 @@ static void waitForInitialZigbeeCoreStartup(void)
...
+    {
+        if (validDelegate)
+        {
+            watchdogDelegate = delegate;
+            shouldKeepDelegate = true;
+            retVal = true;
+        }
+        else
+        {
+            icLogError(LOG_TAG, "%s: invalid watchdog implementation, rejecting", __func__);
```
cleithner-comcast, line 482: minor: lot of new logging, why not use icError?

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -457,6 +454,49 @@ static void waitForInitialZigbeeCoreStartup(void)
...
+    {
+        icLogError(LOG_TAG, "%s: watchdog delegate already set, rejecting", __func__);
+    }
+    mutexUnlock(&watchdogDelegateMtx);
+
+    // Clean up delegate if we didn't keep it
+    if (delegate && !shouldKeepDelegate)
+    {
+        free(delegate);
+    }
```
cleithner-comcast, line 495: This is over complicated. Instead, use g_steal_pointer when moving and free indiscriminantly.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -3470,7 +3529,10 @@ static void zigbeeCoreMonitorFunc(void *arg)
 
     if (zhalHeartbeat(&pid, &zigbeeCoreInitialized) == 0)
     {
-        //WATCHDOG: ZigbeeCore is working, so pet
+        if (watchdogDelegate)
+        {
+            watchdogDelegate->petZhal();
```
cleithner-comcast, line 3534: What's stopping shutdown from freeing watchdogDelegate between the NULL check and here?

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -584,7 +633,17 @@ static void zigbeeSubsystemShutdown(void)
         pthread_join(commFailMonitorThreadId, NULL);
     }
 
-    // WATCHDOG: cleanup/shutdown the watchdog delegate if needed
+    mutexLock(&watchdogDelegateMtx);
+    if (watchdogDelegate)
```
cleithner-comcast, line 637: In general, avoid holding a mutex while calling a callback. You risk a deadlock if the stack calls back into the module. Instead, use the lock to acquire a local pointer to the callback function, then release the lock.

cleithner-comcast (changes requested):
The concurrency looks sketchy. I forsee random use after frees on shutdown in integration test suites.

core/src/subsystems/zigbee/zigbeeSubsystem.c
kfundecmcsa: Well in `zigbeeSubsystemShutdown`, we execute 

```c
    if (zigbeeCoreMonitorTask > 0)
    {
        cancelRepeatingTask(zigbeeCoreMonitorTask);
        zigbeeCoreMonitorTask = 0;
    }
```

before freeing the watchdog delegate. Given this, I don't think there's a race condition here.

cleithner-comcast: Are there any other threads that can reference this delegate or does it only read during the monitor task?

kfundecmcsa: I implemented ref counting on the watchdog delegates to handle this issue. Please take a look and see what you think.

cleithner-comcast (changes requested):
Really close.

tleacmcsa (approved):

PR #68, opened 2025-08-28
docs(architecture): Update architecture docs

tleacmcsa:
Initial cut at Barton architecture documentation cleanup. Plantuml integration added which can be used to build images for documentation, disabled for now since we don't yet have any plantuml files.

Refs: BARTON-255


docs/images/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.10)
+project(barton-docs NONE)
+
+find_program(PLANTUML_EXEC plantuml)
+
+if(NOT PLANTUML_EXEC)
+    message(FATAL_ERROR "plantuml not found in PATH")
+endif()
+
+file(GLOB PUML_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.puml")
```
tleacmcsa, line 10: bah, i wasted so much time trying to find the right diagramming tool for what I wanted to show, but ended up punting and drawing with draw.io.  Plantuml file support is still here but not yet used.  I'm sure I will add some later so I left this in.

kfundecmcsa:
Something to consider: this commit might be better broken into 2 commits - 1 for the build stuff (Dockerfile/CMake stuff) and another for the actual documentation updates. I bring this up because this follows conventional commits a bit more, but mostly because when we make Dockerfile updates, we have to update the docker-builder tag as well. Of course, we can tag this commit, but in the past we've only tagged `build` commits, which I find a bit more intuitive.

tleacmcsa:
> Something to consider: this commit might be better broken into 2 commits - 1 for the build stuff (Dockerfile/CMake stuff) and another for the actual documentation updates. I bring this up because this follows conventional commits a bit more, but mostly because when we make Dockerfile updates, we have to update the docker-builder tag as well. Of course, we can tag this commit, but in the past we've only tagged `build` commits, which I find a bit more intuitive.

I removed the build changes.  Really shouldnt have been in there since I didnt end up using them.  I'll recreate later if needed.

.devcontainer/devcontainer.json
```diff
@@ -115,8 +92,7 @@
                 // Build and Project Management
                 "npclaudiu.vscode-gn",                   // Ninja build system support
 
-                "doi.fileheadercomment",                 // File header comment generator
-                "asciidoctor.asciidoctor-vscode"         // AsciiDoc support
```
kfundecmcsa, line 119: I don't know what this extension is so it doesn't matter to me to remove it. But just curious, why are we removing it here?

kfundecmcsa (dismissed):

docs/images/arch-overview.png
cleithner-comcast, line 1: BartonCore is labeled as Barton Device Service here.

tleacmcsa: thanks, fixed.

PR #71, opened 2025-09-02
Revert "chore: add release action"

cleithner-comcast:
Reverts rdkcentral/BartonCore#70


tleacmcsa (approved):

cleithner-comcast:
> In conventional commits, this should begin with `revert:` - looks like we have a bug in the PR title checker

"Conventional Commits does not make an explicit effort to define revert behavior. Instead we leave it to tooling authors to use the flexibility of types and footers to develop their logic for handling reverts."

https://www.conventionalcommits.org/en/v1.0.0/

I'm pretty sure coggitto supports the default github revert PR format.

kfundecmcsa:
> "Conventional Commits does not make an explicit effort to define revert behavior. Instead we leave it to tooling authors to use the flexibility of types and footers to develop their logic for handling reverts."
> 
> https://www.conventionalcommits.org/en/v1.0.0/
> 
> I'm pretty sure coggitto supports the default github revert PR format.

Our CONTRIBUTING guidelines are more explicit:

https://github.com/rdkcentral/BartonCore/blob/main/CONTRIBUTING.md#reverting-commits

We may want to redo this section then if we want to handle reverts differently. Which I'm fine with by the way, this is our first revert I think, so it's the first time this came up.

kfundecmcsa (approved):

PR #74, opened 2025-09-05
fix: Redact ZHAL network blob from debug prints

tleacmcsa:
The networkConfigData contains the zigbee network key and should not be printed.

Refs: BARTON-291

### DIFF
--- a/libs/zhal/c/src/zhalDataScrubber.c
+++ b/libs/zhal/c/src/zhalDataScrubber.c
@@ -1,0 +1,47 @@
+//------------------------------ tabstop = 4 ----------------------------------
+//
+// Copyright (C) 2025 Comcast Corporation
+//
+// All rights reserved.
+//
+// This software is protected by copyright laws of the United States
+// and of foreign countries. This material may also be protected by
+// patent laws of the United States and of foreign countries.
+//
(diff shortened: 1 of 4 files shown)


kfundecmcsa (approved):

libs/zhal/c/src/zhalDataScrubber.c
```diff
@@ -0,0 +1,48 @@
...
+/*
+ * Created by Raiyan Chowdhury on 2/21/25.
+ */
+
+#include <pthread.h>
+#include <icUtil/regexUtils.h>
+#include "zhalDataScrubber.h"
+
+REGEX_SIMPLE_REPLACER(NETWORK_CONFIG_DATA_REPLACER, "\"networkConfigData\":\"\\([^\"]*\\)\"", NULL, "<REDACTED>");
+static RegexReplacer *SENSITIVE_DATA_REPLACERS[] = { &NETWORK_CONFIG_DATA_REPLACER, NULL };
```
cleithner-comcast, line 33: I know this PR is a port so not asking for changes. However, if the zhal rework is sufficiently large in scope, we should see if glib regex does any of this better/easier.

cleithner-comcast (approved):

PR #87, opened 2025-10-07
build: remove BCORE_PROVIDE_LIBS option

kfundecmcsa:
The BCORE_PROVIDE_LIBS option existed to support an exisiting client of
Barton but is no longer needed, remove it.


tleacmcsa (approved):
Hmm, not sure how this is going to work but i like the idea of getting rid of this option if we can.

cleithner-comcast (approved):

PR #92, opened 2025-10-17
feat: Matter power source cluster and battery attributes

rchowdcmcsa:
Add support for the Matter power source cluster and track changes in the battery percentage. Assumes one power source per device for now.

Known Issues:
- Most of the example apps in the Matter SDK (excluding a few like the all-clusters-app) do not implement the PowerSource cluster (probably for simplicity's sake). If we were to fail the commissioning process whenever we failed to initialize a PowerSource cluster, then we'd fail to commission most of these example apps during testing, so the code was written so that commissioning proceeds regardless.

Refs: BARTON-305

### DIFF
--- a/core/deviceDrivers/matter/MatterDeviceDriver.cpp
+++ b/core/deviceDrivers/matter/MatterDeviceDriver.cpp
@@ -59,6 +59,7 @@
 #include "clusters/BasicInformation.hpp"
 #include "clusters/GeneralDiagnostics.h"
 #include "clusters/OTARequestor.h"
+#include "clusters/PowerSource.h"
 #include "controller/CHIPCluster.h"
 #include "lib/core/CHIPError.h"
 #include "lib/core/DataModelTypes.h"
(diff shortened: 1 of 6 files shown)


core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -495,6 +496,19 @@ std::vector<MatterCluster *> MatterDeviceDriver::GetCommonClustersToSubscribeTo(
         icInfo("OTA cluster not supported on device %s!", deviceId.c_str());
     }
 
+    // TODO: Matter devices can support multiple power sources; we can check which endpoints they're on by querying
+    // the PowerSourceConfiguration cluster. For now, we're just assuming one power source.
```
kfundecmcsa: As a reader, it's not too clear to me why we are assuming one power source. Does Barton only support one power source maybe?

rchowdcmcsa: So far, we've only ever commissioned devices with one power source (as far as I know), like doorbells and lights and sensors, and accounting for more, in my opinion, adds complexity beyond the scope of this effort, which is being called for by a contact sensor story. It'd also make testing more complicated.

I could update the comment to make this clearer..."we're assuming one power source, because so far those are the only kinds of devices we're commissioning." Is that better? I could also just rip off the bandaid and account for multiple right now, but I don't see the point.

tleacmcsa: PowerSource is on the endpoint because each endpoint could be different.  Consider a bridge which has 1 device per endpoint.  Some could be battery powered, other mains powered.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -310,6 +319,15 @@ namespace barton
         uint64_t lastReportCompletedTimestamp = 0;
         std::shared_ptr<chip::Controller::DeviceController> controller = nullptr;
 
+        // Matter 1.4 section 11.7.4.
```
tleacmcsa: see clusters/PowerSource/Enums.h:280

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -143,6 +143,13 @@ namespace barton
          */
         chip::Optional<NetworkUtils::NetworkInterfaceInfo> GetInterfaceInfo() const;
 
+        /**
+         * Check the device's power source type.
+         * TODO: Account for multiple power sources (i.e. PowerSource clusters on multiple endpoints).
```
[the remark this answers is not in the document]
tleacmcsa: The latter is fine.  We can address after SBMD which is coming quite soon.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -292,6 +299,8 @@ namespace barton
 
         SubscriptionIntervalSecs CalculateFinalSubscriptionIntervalSecs();
 
+        // TODO: Account for multiple power sources (i.e. PowerSource clusters on multiple endpoints).
+        uint32_t GetPowerSourceFeatureMap(CHIP_ERROR &error) const;
```
tleacmcsa: this is weirdly inverted from the others.  Should return CHIP_ERROR.

rchowdcmcsa: good catch, done

PR #101, opened 2025-11-03
feat: Matter WiFi signal strength reporting

rchowdcmcsa:
Add support for the wifi diagnostics cluster in the base Matter device driver and track the RSSI value for wifi devices. Calculate link score and link quality from the RSSI. Per requirements, link scores >45 are "good," 30-45 are "fair," and <30 are "poor."

Refs: BARTON-305

### DIFF
--- a/core/deviceDrivers/matter/MatterDeviceDriver.cpp
+++ b/core/deviceDrivers/matter/MatterDeviceDriver.cpp
@@ -113,7 +113,7 @@ static bool getDeviceClassVersion(void *ctx, const char *deviceClass, uint8_t *v
 
 MatterDeviceDriver::MatterDeviceDriver(const char *driverName, const char *deviceClass, uint8_t dcVersion) :
     driver(), deviceClassVersion(dcVersion + deviceModelVersion), otaRequestorEventHandler(*this),
-    generalDiagnosticsEventHandler(*this)
+    generalDiagnosticsEventHandler(*this), wifiDiagnosticsClusterEventHandler(*this)
 {
     driver.driverName = strdup(driverName);
     driver.callbackContext = this;
@@ -463,6 +463,52 @@ void MatterDeviceDriver::ReadResource(std::forward_list<std::promise<bool>> &pro
                                       chip::Messaging::ExchangeManager &exchangeMgr,
                                       const chip::SessionHandle &sessionHandle)
 {
+    icDebug("%s", resource->id);
+
+    if (g_strcmp0(resource->id, COMMON_DEVICE_RESOURCE_FERSSI) == 0 ||
+        g_strcmp0(resource->id, COMMON_DEVICE_RESOURCE_LINK_SCORE) == 0 ||
+        g_strcmp0(resource->id, COMMON_DEVICE_RESOURCE_LINK_QUALITY) == 0)
+    {
+        auto *wifiDiagServer = static_cast<WifiNetworkDiagnostics *>(
+            GetAnyServerById(deviceId, chip::app::Clusters::WiFiNetworkDiagnostics::Id));
+
+        if (wifiDiagServer == nullptr)
+        {
+            icError("WiFiNetworkDiagnostics cluster not present on device %s!", deviceId.c_str());
+            FailOperation(promises);
+            return;
+        }
+
+        promises.emplace_front();
+        auto &readPromise = promises.front();
+        auto readContext = new ClusterReadContext {};
+        readContext->driverContext = &readPromise;
+        readContext->value = value;
+        readContext->resourceId = resource->id;
+
+        // Paired with WifiNetworkDiagnosticsEventHandler::RssiReadComplete()
+        AssociateStoredContext(readContext->driverContext);
+
+        if (!wifiDiagServer->GetRssi(readContext, exchangeMgr, sessionHandle))
+        {
+            AbandonDeviceWork(readPromise);
+            delete readContext;
+        }
+    }
+    else
+    {
+        // If the resource read was not handled above, then let the subclass try to handle it
+        DoReadResource(promises, deviceId, resource, value, exchangeMgr, sessionHandle);
+    }
+}
+
+void MatterDeviceDriver::DoReadResource(std::forward_list<std::promise<bool>> &promises,
+                                        const std::string &deviceId,
+                                        icDeviceResource *resource,
+                                        char **value,
+                                        chip::Messaging::ExchangeManager &exchangeMgr,
+                                        const chip::SessionHandle &sessionHandle)
+{
     icDebug("Unimplemented");
     FailOperation(promises);
 }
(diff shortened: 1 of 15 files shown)


core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -1402,6 +1496,89 @@ void MatterDeviceDriver::OtaRequestorEventHandler::OnVersionApplied(OTARequestor
...
+    // percentage close to 100. A normalizedRSSI close to 1 (bad) should cause the second subtraction operand being
+    // close to 100, leading to a percentage close to 0.
+    uint8_t percentage = 100 - (100 * normalizedRSSI);
+
+    return percentage;
+}
+
+std::string MatterDeviceDriver::GetLinkQuality(uint8_t linkScore)
+{
+    if (linkScore > 45)
```
cleithner-comcast: These values used to be based on properties when we did this for zigbee devices. I am certain that if they are a matter feature we support, someone will want them to be properties again.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -1402,6 +1496,89 @@ void MatterDeviceDriver::OtaRequestorEventHandler::OnVersionApplied(OTARequestor
...
+    {
+        rssiStr = g_strdup_printf("%" PRId8, *rssi);
+        linkScore = GetLinkScore(*rssi);
+        linkScoreStr = g_strdup_printf("%" PRIu8, linkScore);
+        linkQuality = GetLinkQuality(linkScore);
+    }
+
+    updateResource(deviceUuid.c_str(), nullptr, COMMON_DEVICE_RESOURCE_FERSSI, rssiStr, nullptr);
+    updateResource(deviceUuid.c_str(), nullptr, COMMON_DEVICE_RESOURCE_LINK_SCORE, linkScoreStr, nullptr);
+    updateResource(deviceUuid.c_str(), nullptr, COMMON_DEVICE_RESOURCE_LINK_QUALITY, linkQuality.c_str(), nullptr);
```
cleithner-comcast: Where's the valve for this firehose? Attribute reporting configuration, resource event emission throttle configuration, or at the client level (i.e. every client implements their own event debounce)?

cleithner-comcast (dismissed):

core/deviceDrivers/matter/MatterDeviceDriver.cpp
rchowdcmcsa: I gave the spec another read, and the RSSI attribute actually has the Changes Omitted quality, which "MAY be given to attribute data that is deemed to have a high rate of change or where changes are not meaningful or too large to convey as part of Subscribe interaction [...] Attribute data with this quality SHALL support Read Access, but SHALL NOT have delta changes published as part of a Subscribe interaction." So, RSSI changes are not even included in subscription reports actually (I remember this happening with the chime way back when, but I didn't know why at the time).

We would either have to periodically query the device for its RSSI value or only read it when we need it (not sure what entails "needing" it, but we would probably just write the APIs to read it and let other people worry about invoking them at the right time).

core/deviceDrivers/matter/clusters/WifiNetworkDiagnostics.cpp
```diff
@@ -43,69 +43,56 @@ extern "C" {
...
 
-        using namespace chip::app::Clusters::WiFiNetworkDiagnostics;
-        using TypeInfo = Attributes::Rssi::TypeInfo;
+    using namespace chip::app::Clusters::WiFiNetworkDiagnostics;
+    using TypeInfo = Attributes::Rssi::TypeInfo;
 
-        chip::app::ConcreteAttributePath path(endpointId,
-                                              chip::app::Clusters::WiFiNetworkDiagnostics::Id,
-                                              chip::app::Clusters::WiFiNetworkDiagnostics::Attributes::Rssi::Id);
+    // Per Matter 1.4 11.15.6, the RSSI attribute is not reportable, so its value is not updated in the cache via
```
tleacmcsa: I can't remember if you confirmed whether or not rssi is in the cache at all.  If not, please update the comment.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
tleacmcsa: The solution could possibly be common, but we dont know yet and with SBMD coming we dont know what a common spot would look like right now.

core/deviceDrivers/matter/MatterDeviceDriver.h
```diff
@@ -202,14 +203,23 @@ namespace barton
...
 
+        void ReadCommonResource(std::forward_list<std::promise<bool>> &promises,
+                                const std::string &deviceId,
+                                icDeviceResource *resource,
+                                char **value,
+                                chip::Messaging::ExchangeManager &exchangeMgr,
+                                const chip::SessionHandle &sessionHandle);
+
+        /**
+         * For device-specific resources.
```
tleacmcsa: This comment doesnt match the implementation

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -455,6 +455,54 @@ void MatterDeviceDriver::SetTampered(const std::string &deviceId, bool tampered)
     icDebug("Unimplemented");
 }
 
+void MatterDeviceDriver::ReadCommonResource(std::forward_list<std::promise<bool>> &promises,
```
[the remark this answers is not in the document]
tleacmcsa: One way to handle that is for the base class/driver's ReadResource to call ReadDeviceResource if the leaf class doesn't handle it.  Maybe by a virtual method DoReadResource which is what the leaf class would override instead.

The base driver's read resource functionality may handle the request internally (if its a resource on the device or one of its endpoints and its a common one), or if it doesnt handle it, let the leaf driver do it.  In all cases, the leaf driver must be able to overwrite the functionality.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -1292,6 +1373,62 @@ void MatterDeviceDriver::OtaRequestorEventHandler::OnVersionApplied(OTARequestor
...
+                *readContext->value = g_strdup(rssiStr);
+            }
+            else if (g_strcmp0(readContext->resourceId, COMMON_DEVICE_RESOURCE_LINK_SCORE) == 0)
+            {
+                *readContext->value = g_strdup(linkScoreStr);
+            }
+            else if (g_strcmp0(readContext->resourceId, COMMON_DEVICE_RESOURCE_LINK_QUALITY) == 0)
+            {
+                *readContext->value = g_strdup(linkQuality);
+            }
```
tleacmcsa: should we log the unexpected final 'else' ?

rchowdcmcsa: Yeah, good idea. Added it.

core/deviceDrivers/matter/clusters/WifiNetworkDiagnostics.cpp
```diff
@@ -43,69 +43,56 @@ extern "C" {
...
-        using TypeInfo = Attributes::Rssi::TypeInfo;
+    using namespace chip::app::Clusters::WiFiNetworkDiagnostics;
+    using TypeInfo = Attributes::Rssi::TypeInfo;
 
-        chip::app::ConcreteAttributePath path(endpointId,
-                                              chip::app::Clusters::WiFiNetworkDiagnostics::Id,
-                                              chip::app::Clusters::WiFiNetworkDiagnostics::Attributes::Rssi::Id);
+    // Per Matter 1.4 11.15.6, the RSSI attribute is not reportable, so its value is not populated in the cache via
+    // subscription reports. Therefore, we must read it on-demand.
+    auto rssiReadContext = new OnDemandReadContext(context, eventHandler, deviceId);
```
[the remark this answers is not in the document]
rchowdcmcsa: I think I've come up with an answer -- please standby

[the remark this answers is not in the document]
rchowdcmcsa: done

PR #102, opened 2025-11-13
refactor(matter): use wildcard subscription cache

tleacmcsa:
Replaces Zigbee-inspired phased Matter discovery with a single wildcard
subscription + local ClusterStateCache via DeviceDataCache. Eliminates
burst/adhoc attribute reads at commissioning and runtime; all local reads
now served from an always-on subscription-fed cache.

Motivation:
- Reduce network round trips and commissioning latency
- Provide consistent, immediate snapshot + incremental updates
- Simplify driver claiming and resource creation

Key Changes:
- Added DeviceDataCache (wildcard subscribe, ClusterStateCache wrapper, JSON export, endpoint & network accessors)
- CommissioningOrchestrator creates DeviceDataCache; driver chosen by inspecting cached descriptor/device types
- MatterDeviceDriver: new ClaimDevice(const DeviceDataCache*), ConfigureDevice/SynchronizeDevice after cache init
- Introduced DoRegisterResources / DoSynchronizeDevice hooks (Matter thread context)
- Removed DiscoveredDeviceDetailsStore and legacy initialResourceValues / subscription configuration path
- Removed SubscribeInteraction class and tests
- Refactored Light, DoorLock, WindowCovering drivers and related clusters
- Replaced subscriptionMap with deviceDataCaches (mutex guarded)
- Added jsontlv dependency for TLV→JSON cluster state logging
- Added DeviceDataCacheTest covering construction, accessors, handler lifecycle

Behavioral Impact:
- Commissioning waits for initial subscription report; subsequent reads are cache hits
- Reprocessing the cache after resources are created currently result in additional resource updated events
  and initial resource values are briefly empty

Risks / Known Issues:
- InitializeDeviceDataCacheIfRequired blocks on future without timeout (possible stall) [TODO]
- Subscription interval logic hardcoded (3360s ceiling heuristic); removed earlier interval/backoff tests
- Event handling placeholder (OnEventData) not yet fully implemented

Follow-ups (recommended):
1. Add timeout/non-blocking start for DeviceDataCache
2. Reintroduce tests for subscription interval & backoff logic
3. Resource initialization happens after creation with events
4. Driver-specific subscription interval (remove hardcoded 3360s)
5. Guard JSON export size or provide sampling to avoid log spam

Testing:
- DeviceDataCacheTest exercises construction & basic behaviors
- Manual commissioning should show immediate snapshot in cache debug JSON
- Driver claiming validated via DeviceTypeList decode paths

Migration Notes:
- External code using DiscoveredDeviceDetails must switch to DeviceDataCache (via driver GetDeviceDataCache)
- endpointProfileMap now always null; use DeviceDataCache::GetEndpointIds() to enumerate endpoints

Refs: BARTON-316

### DIFF
--- a/core/src/subsystems/matter/DeviceDataCache.cpp
+++ b/core/src/subsystems/matter/DeviceDataCache.cpp
@@ -46,11 +46,8 @@ using namespace barton::Subsystem;
 DeviceDataCache::~DeviceDataCache()
 {
     icDebug();
-    // Explicitly destroy in the correct order to avoid use-after-free
-    // readClient must be destroyed before clusterStateCacheCallback since it calls OnDeallocatePaths
-    readClient.reset();
-    clusterStateCache.reset();
-    clusterStateCacheCallback.reset();
+    // readClient and clusterStateCache unique_ptrs will be automatically cleaned up
+    // No need for ScheduleWork since they handle their own destruction safely
 }
 
 std::future<bool> DeviceDataCache::Start()
@@ -414,32 +411,6 @@ void DeviceDataCache::OnReportBegin()
     icDebug();
 }
 
-void DeviceDataCache::OnEventData(const chip::app::EventHeader &aEventHeader,
-                                  chip::TLV::TLVReader *apData,
-                                  const chip::app::StatusIB *apStatus)
-{
-    auto key = std::make_tuple(deviceUuid, aEventHeader.mPath.mEndpointId, aEventHeader.mPath.mClusterId);
-    auto it = clusterCallbacks.find(key);
-    chip::app::ClusterStateCache::Callback *callback = (it != clusterCallbacks.end()) ? it->second : nullptr;
-    if (callback != nullptr)
-    {
-        callback->OnEventData(aEventHeader, apData, apStatus);
-    }
-}
-
-void DeviceDataCache::OnAttributeData(const chip::app::ConcreteDataAttributePath &aPath,
-                                      chip::TLV::TLVReader *apData,
-                                      const chip::app::StatusIB &aStatus)
-{
-    auto key = std::make_tuple(deviceUuid, aPath.mEndpointId, aPath.mClusterId);
-    auto it = clusterCallbacks.find(key);
-    chip::app::ClusterStateCache::Callback *callback = (it != clusterCallbacks.end()) ? it->second : nullptr;
-    if (callback != nullptr && aStatus.IsSuccess())
-    {
-        callback->OnAttributeChanged(clusterStateCache.get(), aPath);
-    }
-}
-
 Json::Value DeviceDataCache::GetEndpointAsJson(chip::EndpointId endpointId)
 {
     Json::Value endpointJson(Json::objectValue);
@@ -649,8 +620,7 @@ void DeviceDataCache::OnDeviceConnected(chip::Messaging::ExchangeManager &exchan
 {
     icDebug();
 
-    clusterStateCacheCallback = std::make_unique<ClusterStateCacheCallback>(this);
-    clusterStateCache = std::make_unique<chip::app::ClusterStateCache>(*clusterStateCacheCallback);
+    clusterStateCache = std::make_unique<chip::app::ClusterStateCache>(*this);
     readClient =
         std::make_unique<chip::app::ReadClient>(chip::app::InteractionModelEngine::GetInstance(),
                                                 chip::app::InteractionModelEngine::GetInstance()->GetExchangeManager(),
--- a/core/src/subsystems/matter/DeviceDataCache.h
+++ b/core/src/subsystems/matter/DeviceDataCache.h
@@ -41,7 +41,7 @@ using namespace barton::Subsystem::Matter;
 namespace barton
 {
 
-    class DeviceDataCache
+    class DeviceDataCache : public chip::app::ClusterStateCache::Callback
     {
     public:
         DeviceDataCache(const std::string &deviceUuid,
@@ -152,53 +152,6 @@ namespace barton
         CHIP_ERROR RegenerateAttributeReport();
 
     private:
-        /**
-         * Inner callback class that implements ClusterStateCache::Callback interface.
-         * Delegates all callbacks back to the outer DeviceDataCache instance.
-         */
-        class ClusterStateCacheCallback : public chip::app::ClusterStateCache::Callback
-        {
-        public:
-            explicit ClusterStateCacheCallback(DeviceDataCache *parent) : parent(parent) {}
-
-            void NotifySubscriptionStillActive(const chip::app::ReadClient &apReadClient) override {}
-
-            void OnReportBegin() override { parent->OnReportBegin(); }
-
-            void OnReportEnd() override { parent->OnReportEnd(); }
-
-            void OnEventData(const chip::app::EventHeader &aEventHeader,
-                             chip::TLV::TLVReader *apData,
-                             const chip::app::StatusIB *apStatus) override
-            {
-                parent->OnEventData(aEventHeader, apData, apStatus);
-            }
-
-            void OnAttributeData(const chip::app::ConcreteDataAttributePath &aPath,
-                                 chip::TLV::TLVReader *apData,
-                                 const chip::app::StatusIB &aStatus) override
-            {
-                parent->OnAttributeData(aPath, apData, aStatus);
-            }
-
-            void OnSubscriptionEstablished(chip::SubscriptionId aSubscriptionId) override
-            {
-                parent->OnSubscriptionEstablished(aSubscriptionId);
-            }
-
-            void OnError(CHIP_ERROR aError) override { parent->OnError(aError); }
-
-            void OnDeallocatePaths(chip::app::ReadPrepareParams &&aReadPrepareParams) override
-            {
-                parent->OnDeallocatePaths(std::move(aReadPrepareParams));
-            }
-
-            void OnDone(chip::app::ReadClient *apReadClient) override { parent->OnDone(apReadClient); }
-
-        private:
-            DeviceDataCache *parent;
-        };
-
         static void OnDeviceConnectedCallback(void *context,
                                               chip::Messaging::ExchangeManager &exchangeMgr,
                                               const chip::SessionHandle &sessionHandle);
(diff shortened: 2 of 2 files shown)


cleithner-comcast:
Seems to be a failure with the integration test suite. Tests appeared to pass but script had non-zero exit code. You might run locally and see if there's a leak or something not reported in the console.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -0,0 +1,299 @@
...
+        /**
+         * Get a string attribute value from the cache.
+         *
+         * @param endpointId The endpoint ID.
+         * @param clusterId The cluster ID.
+         * @param attributeId The attribute ID.
+         * @return The string value, or empty string if not found or on error.
+         */
+        std::string
+        GetStringAttribute(chip::EndpointId endpointId, chip::ClusterId clusterId, chip::AttributeId attributeId) const;
```
cleithner-comcast: I think an error enum might be good for reads. At the very least, a driver would wanna report (log, telemetry) if an expected cluster or attribute is missing vs the device wasn't found cause its offline vs a network failure, et al.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -0,0 +1,299 @@
...
+         * @return True if the endpoint has the server cluster, false otherwise.
+         */
+        bool EndpointHasServerCluster(chip::EndpointId endpointId, chip::ClusterId clusterId) const;
+
+        /**
+         * Get a complete JSON representation of all cached device data.
+         *
+         * @return Json::Value object containing all endpoints, clusters, and attributes.
+         */
+        Json::Value GetCacheAsJson();
```
cleithner-comcast: Does anyone actually need this? I can't think of a use case for the driver and I'd rather abstract the message payload and its typing.

core/src/subsystems/matter/DeviceDataCache.cpp
```diff
@@ -0,0 +1,536 @@
...
+    };
+
+    return interfaceInfo.Value().networkType;
+}
+
+chip::Optional<NetworkUtils::NetworkInterfaceInfo> DeviceDataCache::GetInterfaceInfo() const
+{
+    using TypeInfo = chip::app::Clusters::GeneralDiagnostics::Attributes::NetworkInterfaces::TypeInfo;
+    TypeInfo::DecodableType value;
+    CHIP_ERROR error = clusterStateCache->Get<TypeInfo>({chip::kRootEndpointId,
```
cleithner-comcast: You might take a comb to spacing in this change. Normally I'm not too bothered by many lines of code with no spacing but when they're particularly long and even have statements that span multiple lines AND there's no spacing, my eyes start to cross a bit.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -151,9 +148,80 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
+            continue;
+        }
+
+        // Check if this endpoint has the Descriptor cluster
+        if (!deviceDataCache->EndpointHasServerCluster(endpointId, chip::app::Clusters::Descriptor::Id))
+        {
+            continue;
+        }
+
+        // Read the DeviceTypeList attribute from the Descriptor cluster
```
cleithner-comcast: This feels like it should live in DeviceDataCache to me. Ideally, a driver (at least the common parent MatterDeviceDriver) should really never have to go querying the clusterstatecache directly and decoding the TLV.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -547,13 +482,50 @@ bool MatterDeviceDriver::CreateResources(icDevice *device, icInitialResourceValu
...
+
+                DoSynchronizeDevice(promises, deviceId, exchangeMgr, sessionHandle);
+            },
+            MATTER_ASYNC_SYNCHRONIZE_DEVICE_TIMEOUT_SECS))
+    {
+        icError("Failed to sync device %s", device->uuid);
+    }
+}
+
+void MatterDeviceDriver::DoSynchronizeDevice(std::forward_list<std::promise<bool>> &promises,
```
cleithner-comcast: Should this just be an invariant success? With the cluster state cache, the concept of synchronize device is mostly irrelevant for matter devices, right?

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -595,30 +567,67 @@ void MatterDeviceDriver::ExecuteResource(std::forward_list<std::promise<bool>> &
...
 
-    DiscoveredDeviceDetailsStore::Instance().Put(device->uuid, details);
-    result = true;
+    createDeviceResource(device,
+                         COMMON_DEVICE_RESOURCE_SERIAL_NUMBER,
+                         serialNumber,
+                         RESOURCE_TYPE_SERIAL_NUMBER,
+                         RESOURCE_MODE_READABLE,
+                         CACHING_POLICY_ALWAYS);
+    free(serialNumber);
```
cleithner-comcast: consider g_autofree

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -821,39 +830,32 @@ bool MatterDeviceDriver::ConnectAndExecute(const std::string &deviceId, connect_
     return !abort;
 }
 
-static bool configureDevice(void *self, icDevice *device, DeviceDescriptor *descriptor)
+bool MatterDeviceDriver::InitializeDeviceDataCacheIfRequired(const std::string &deviceUuid)
```
cleithner-comcast: It's a little bit weird that there are two areas where a driver gets its device data cache. Is there not a way for the driver to always make it and the commissioning orchestrator just ask the driver for it?

It's also kind of weird that when the driver makes its device data cache, it is lazily created instead of as a construction invariant.

core/src/subsystems/matter/DeviceDataCache.h
tleacmcsa: Agree.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -0,0 +1,299 @@
...
+        // Some simple accessors for common Basic Information Cluster attributes
+        std::string GetVendorName() const;
+        std::string GetProductName() const;
+        std::string GetHardwareVersionString() const;
+        std::string GetSoftwareVersionString() const;
+        std::optional<std::string> GetSerialNumber() const;
+
+        // Some common accessors for General Diagnostics Cluster attributes
+        std::string GetMacAddress() const;
+        std::string GetNetworkType() const;
```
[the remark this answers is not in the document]
tleacmcsa: The point of these helpers are to map from Matter types into strings, which is how we represent all resource values today.  If I expose Spans as the return type only for every caller to convert to string, that isn't very nice.  These are optional helpers and callers should be able to get at the TLV in a future change which is likely required for SBMDs anyway.

core/src/subsystems/matter/DeviceDataCache.h
tleacmcsa: At the moment this is used for debugging and also to support the first iteration of SBMD that would (only at first) operate on JSON.  Likely to go away later if we no longer need it.

core/src/subsystems/matter/CommissioningOrchestrator.cpp
```diff
@@ -159,31 +159,33 @@ namespace barton
 
         if (commissioningStatus == DiscoveryCompleted)
         {
-            auto matterDetails = discoverer.GetDetails();
-
-            MatterDeviceDriver *driver = MatterDriverFactory::Instance().GetDriver(matterDetails.get());
+            // GetDriver needs to inspect the cache to determine driver type
+            MatterDeviceDriver *driver = MatterDriverFactory::Instance().GetDriver(deviceDataCache.get());
```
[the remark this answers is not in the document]
tleacmcsa: Meh, this is used in exactly one place (here), but I'll take a look.  Also, not new or related to this change.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
tleacmcsa: Yeah, we already have accessors in DeviceDataCache, so it probably makes sense to move there.

Funny thing about drivers and TLV: this is exactly where I'm taking SBMD.  Only the SBMD knows what clusters to use and would be handling the TLV codec, but I have a plan for that.  Building knowledge of all clusters accessible by drivers into Barton directly will artificially limit SBMDs.

core/src/subsystems/matter/DeviceDataCache.h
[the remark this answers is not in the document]
cleithner-comcast: Hmm I was of the mindset that it would be likely that a driver would want to do logic on some of these (eg. do something different based on network type) but you believe most of the time drivers will just want strings to shove into resources?

core/src/subsystems/matter/DeviceDataCache.h
cleithner-comcast: Ack, you may indicate as much since there may be custom matter drivers in development and removing this later would break them. At least with documentation we can curb breakage risk.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
cleithner-comcast: Ack

core/src/subsystems/matter/DeviceDataCache.h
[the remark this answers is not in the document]
tleacmcsa: The base matter driver would take care of the boring and basic mapping of the common attributes to these string resources.  Concrete drivers would very likely not use these at all and instead interact with the real Matter types and do whatever they need to map those into attributes, commands, etc.  I see there are not good examples of that in this PR since I was building up to SBMD, so I'll show that here by going ahead and addressing the "initial resource values" gap.

core/src/subsystems/matter/DeviceDataCache.h
tleacmcsa: Ack.

core/src/subsystems/matter/DeviceDataCache.h
```diff
@@ -0,0 +1,299 @@
...
+         * @return Json::Value object containing all endpoints, clusters, and attributes.
+         */
+        Json::Value GetCacheAsJson();
+
+        /**
+         * Get the cluster state cache.
+         *
+         * @return Pointer to the cluster state cache, or nullptr if not initialized.
+         */
+        chip::app::ClusterStateCache *GetClusterStateCache() const { return clusterStateCache.get(); }
```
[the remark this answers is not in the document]
tleacmcsa: Got rid of it.

core/deviceDrivers/matter/MatterDoorLockDeviceDriver.cpp
```diff
@@ -194,19 +152,6 @@ bool MatterDoorLockDeviceDriver::RegisterResources(icDevice *device, icInitialRe
...
-        initialResourceValues, DOOR_LOCK_ENDPOINT, DOORLOCK_PROFILE_RESOURCE_LOCKED);
-    if (initialValue)
-    {
-        resource->value = strdup(initialResourceValuesGetEndpointValue(
-            initialResourceValues, DOOR_LOCK_ENDPOINT, DOORLOCK_PROFILE_RESOURCE_LOCKED));
-    }
-    else
-    {
-        result = false;
-    }
```
cleithner-comcast: Why do we not populate locked initially anymore?

mhughesacn:
Doesn't need to block but strictly the complaint is correct i.e. json files should not have a comment.

cleithner-comcast (approved):

PR #109, opened 2025-12-04
fix: Create networkType resource for zigbee devices

Lakshmi97-velampati:
This change creates the networkType resource for all the zigbee devices if it 
is not exists already.As part of this update, the zigbee common device driver
schema version has been incremented to reflect the change in resource
initialization behavior.

Refs: XHCPE-1991


tleacmcsa:
@Lakshmi97-velampati please update the title and commit message.  This change is not sensor specific, rather it applies to all zigbee devices.  Also please give details on the schema version update in the commit message.

PR #119, opened 2026-02-05
feat(matter): add SbmdScript interface and QuickJS implementation

tleacmcsa:
Add the scripting layer for SBMD that enables dynamic behavior in device drivers through JavaScript.

- SbmdScript.h: Abstract interface for script execution
- QuickJsScript: QuickJS-based implementation of the script interface
- Provides TLV-to-JavaScript and JavaScript-to-TLV conversion
- Supports attribute reads, writes, commands, and event handling
- QuickJsScriptTest.cpp: Unit tests for QuickJS script execution
- Add quickjs to Dockerfile

### DIFF
--- a/core/deviceDrivers/matter/sbmd/script/QuickJsScript.cpp
+++ b/core/deviceDrivers/matter/sbmd/script/QuickJsScript.cpp
@@ -25,6 +25,7 @@
 // Created by tlea on 12/5/25
 //
 
+#include "json/writer.h"
 #define LOG_TAG "QuickJsScript"
 
 #include "QuickJsScript.h"
@@ -39,6 +40,7 @@
 #include <mutex>
 #include <platform/CHIPDeviceLayer.h>
 #include <sstream>
+#include <unordered_map>
 
 extern "C" {
 #include <icLog/logging.h>
@@ -275,7 +277,14 @@ namespace barton
                 {
                     if (elem.isIntegral())
                     {
-                        bytes.push_back(static_cast<uint8_t>(elem.asInt()));
+                        int value = elem.asInt();
+                        if (value < 0 || value > 255)
+                        {
+                            icLogError(
+                                LOG_TAG, "Invalid value in byte array for argument '%s': %d", arg.name.c_str(), value);
+                            return Json::writeString(Json::StreamWriterBuilder(), tlvJson);
+                        }
+                        bytes.push_back(static_cast<uint8_t>(value));
                     }
                     else
                     {
--- a/core/deviceDrivers/matter/sbmd/SbmdSpec.h
+++ b/core/deviceDrivers/matter/sbmd/SbmdSpec.h
@@ -50,17 +50,20 @@ namespace barton
         // Equality operator for map key usage
         bool operator==(const SbmdAttribute &other) const
         {
-            return clusterId == other.clusterId && attributeId == other.attributeId;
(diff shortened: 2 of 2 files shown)


core/deviceDrivers/matter/sbmd/script/QuickJsScript.h
```diff
@@ -0,0 +1,240 @@
...
+         * {"output" : <Barton string representation of the command response>}
+         *
+         * @see SbmdScript::MapCommandExecuteResponse
+         */
+        bool MapCommandExecuteResponse(const SbmdCommand &commandInfo,
+                                       chip::TLV::TLVReader &reader,
+                                       std::string &outValue) override;
+
+    private:
+        explicit QuickJsScript(const std::string &deviceId, JSRuntime *runtime, JSContext *ctx);
```
cleithner-comcast: What is explicit doing here? I thunk it's just for single argument constructors to prevent implicit conversion?

cleithner-comcast (commented):
Have to break, will come back. Reviewed everything except QuickJsScript.cpp.

core/deviceDrivers/matter/sbmd/SbmdParser.cpp
```diff
@@ -646,12 +646,12 @@ uint32_t SbmdParser::ParseHexOrDecimal(const std::string &value)
     }
     catch (const std::invalid_argument &e)
     {
-        icError(logFmt("Invalid numeric value '%s': %s"), value.c_str(), e.what());
+        icLogError(LOG_TAG, "(%s): Invalid numeric value '%s': %s", __func__, value.c_str(), e.what());
```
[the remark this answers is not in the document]
tleacmcsa: Yeah, I know.  "Someone" is stuck on that.  I will fix them all at the end of all PRs.  It's annoying.

core/deviceDrivers/matter/sbmd/script/QuickJsScript.h
tleacmcsa: explicit prevents implicit conversions when constructing objects. For a constructor with multiple parameters like this one, it's technically redundant since C++ only performs implicit conversions for single-argument constructors.

However, it's still good practice here because:

Future-proofing - if parameters later get default values, making it a single-argument constructor, the explicit protection is already in place
Intent clarity - signals this constructor should only be called deliberately via the factory method Create()
Consistency - many style guides recommend explicit on all constructors
That said, since this is a private constructor only called by the static Create() method, the explicit keyword has no practical effect and could be removed without changing behavior.

core/deviceDrivers/matter/sbmd/script/QuickJsScript.cpp
```diff
@@ -0,0 +1,854 @@
...
+
+    // Wrap the script body in a function and execute it
+    std::string wrappedScript = "(function() { " + script + " })()";
+
+    icLogDebug(LOG_TAG, "Executing script: %s", wrappedScript.c_str());
+
+    // Execute the script
+    JSValue scriptResult = JS_Eval(ctx, wrappedScript.c_str(), wrappedScript.length(),
+                                   "<sbmd_script>", JS_EVAL_TYPE_GLOBAL);
+    if (JS_IsException(scriptResult))
```
cleithner-comcast: Several duplicate blocks of this. Consider isolating to a common helper function.

PR #128, opened 2026-02-10
feat(matter): add SBMD specs dir init param property

tleacmcsa:
The SBMD system needs to know where to find the specifications. This change adds a new Barton initialize param property for this directory.


tleacmcsa:
> > > The PR title is a bit of a misgnomer. We're adding an initialize param property, but just the word "property" to me means a Barton property. On that note, presumably we would want to support runtime installation of new specs, so wouldn't a property be more appropriate than an initial param?
> > 
> > 
> > Fixed PR title/body.
> > This path needs to be set very early so we can load the drivers. I think init param is right.
> 
> Would it not be appropriate to be time agnostic and capable of "reloading" drivers any time the property is set? I'm just thinking ahead here. It would be somewhat annoying later to decide we want to support side channel distribution of specs and thus need a property anyway, deprecating the param in the API layer and eventually removing it.

We can reload drivers in the future if we want.  but i dont see a need to be able to change the dir dynamically.

cleithner-comcast (approved):

PR #197, opened 2026-03-30
docs: propose camera architecture redesign

kfundecmcsa:
docs: propose camera architecture redesign

Add openspec change for a protocol-agnostic camera stream
contract to support Matter 1.5 WebRTC cameras alongside
existing OpenHome cameras. Two-layer endpoint architecture:

- Abstract camera endpoint (stream, streamStatus) drives
  stream lifecycle via events with per-stream correlation
- Protocol-specific endpoints (webrtc, openhome) carry
  signaling details per technology
- Stream coordination reuses existing updateResource()
  metadata mechanism (cJSON convention, not typed)

Design decisions cover endpoint structure, event metadata
schema, events-as-source-of-truth, driver-determined
protocol selection, legacy resource cleanup, and Matter 1.5
WebRTC transport direction.

Implementation is scoped as two follow-on tasks, each
going through its own openspec explore/propose/apply cycle:

1. Camera stream contract + OpenHome adapter - define
   the contract and prove it e2e on existing hardware
2. Matter camera support via WebRTC - new driver, cluster
   server hosting, signaling, and media stack (depends
   on task 1)

BARTON-334, BARTON-354


tleacmcsa (changes requested):
I didn't get too far in this review since there is a much larger and higher level discussion about how to model devices that needs to be had.

cleithner-comcast:
Should this be draft?

kfundecmcsa:
> Should this be draft?

probably shoulda, but just put up another commit up for review

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
+_These tasks represent follow-on implementation work. Each is expected to go through its own explore → propose → apply cycle._
+
+## 1. Camera session contract and OpenHome adapter
+
+**Capabilities**: `camera-session-contract`, `openhome-session-adapter`
+
+Define the protocol-agnostic camera session contract and prove it end-to-end by adapting the existing OpenHome camera driver. This is the foundational work — it lands the architecture in code using existing hardware with no new external dependencies.
+
+Scope includes:
+- Session lifecycle resources (`startSession`, `stopSession`, `sessionStatus`) and metadata constants (`sessionId`, `technology`, `nextAction`, status values) in `commonDeviceDefs.h`
```
cleithner-comcast: Not sure stopSession is necessary. I also don't think the primary interface for a user should be startSession. I would expect to execute something like "startVideoStream" resource and the session info is emergent.

Sessions should be self managed by the driver (camera parent class driver?) and have its own timeout and cleanup of session info.

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+## 2. Matter camera support via WebRTC
+
+**Capabilities**: `webrtc-endpoint-profile`
+**Depends on**: Task 1 (camera session contract)
+
+Implement the Matter 1.5 WebRTC camera driver and signaling flow on top of the session contract from Task 1.
+
+Scope includes:
+- `webrtc` endpoint profile with `peerSdp`, `cameraSdp`, `peerIceCandidates`, `cameraIceCandidates` resources and `WEBRTC_PROFILE_*` constants
+- Matter camera driver (native C/C++) — including SBMD feasibility evaluation to determine whether SBMD can express stateful, bidirectional WebRTC signaling or whether a native driver is required
```
cleithner-comcast: This is worded weirdly. It should be SBMD, but the wrench is you may want a common parent class. Which means a new problem needs to be solved: how do you define a Matter Driver with SBMD composition but also could have one ore more additional class dependencies?

Another option maybe is that an extantiation of the  session management class is owned by the openhome driver but also has js bindings and is available in the context to the SBMD.

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+- Cluster server hosting pattern (`WebRtcTransportRequestorServer`) — first server-side cluster in Barton's Matter subsystem
+- SDP offer/answer exchange and ICE candidate negotiation
+- WebRTC media stack integration (libdatachannel or alternative)
+- Thread marshalling from CHIP event loop to GLib main loop
```
cleithner-comcast: Huh?

cleithner-comcast (changes requested):

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+  |                             |── createMediaTunnel() ─► |
+  |                             |                          |
+  | ◄── sessionStatus event ──  |                          |
+  |   value: "connected"        |                          |
+  |   meta: {sessionId: "abc", technology: "openhome",|
+  |          nextAction: "/dev/ep/openhome/r/tunnelUrl"}
+  |                             |                          |
+  | ◄── tunnelUrl event ──────  |                          |
+  |   value: "rtsp://..."       |                          |
+```
```
rchowdcmcsa, line 188: So these session flows document the `startSession` flow, but I don't see anything similar for `stopSession`. What would the `sessionStatus` value and metadata look like during that flow(s)? What happens if `stopSession` is called with no active session? What happens if it's called mid-signaling? What cleanup (if any) is involved?

I see the other comment questioning whether or not `stopSession` is even necessary. I don't know if it is or not, but that aside, it just feels like it was glossed over in the proposal so far.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+  |   (startSession, {})  ──►   |                          |
+  |                             |── createMediaTunnel() ─► |
+  |                             |                          |
+  | ◄── sessionStatus event ──  |                          |
+  |   value: "connected"        |                          |
+  |   meta: {sessionId: "abc", technology: "openhome",|
+  |          nextAction: "/dev/ep/openhome/r/tunnelUrl"}
+  |                             |                          |
+  | ◄── tunnelUrl event ──────  |                          |
+  |   value: "rtsp://..."       |                          |
```
rchowdcmcsa, line 209: What happens if a client calls `startSession` when a session is already active?

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Risks / Trade-offs
+
+- **[Metadata is untyped]** → Mitigation: Well-documented JSON schema, string constants in `commonDeviceDefs.h`, and integration tests that verify metadata content. Strongly typed metadata is a future capability.
+- **[SBMD may not support the session pattern]** → Mitigation: SBMD feasibility for Matter cameras is explicitly out of scope. The Matter camera driver may need to be a native C/C++ driver rather than SBMD. This is evaluated in the webrtc-endpoint-profile task.
+- **[Parallel sessions deferred]** → Mitigation: The contract is designed for parallel sessions (sessionId in metadata, events as source of truth). Only the implementation is single-session MVP. No contract changes needed later.
+- **[OpenHome adapter complexity]** → Mitigation: The OpenHome driver already implements `createMediaTunnel`. The adapter wraps this in the session contract — `startSession` calls `createMediaTunnel` internally and emits `sessionStatus` events. Incremental, low-risk change.
```
rchowdcmcsa: It says elsewhere that existing OpenHome camera clients call `createMediaTunnel` directly so that the old path continues to work unchanged. What happens if both paths are exercised? Can that even happen?

EDIT: This highlighted a lot of lines. I'm only concerned with line 177 here.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+  "sessionId": "uuid-string",
+  "technology": "webrtc" | "openhome",
+  "nextAction": "/deviceId/ep/webrtc/r/peerSdp"
+}
+```
+
+**Constants** (added to `commonDeviceDefs.h`):
+```c
+#define CAMERA_SESSION_METADATA_SESSION_ID   "sessionId"
+#define CAMERA_SESSION_METADATA_TECHNOLOGY   "technology"
```
rchowdcmcsa: We should probably define constants for metadata values as well, e.g. a constant for the technology value "webrtc" and so on.

openspec/changes/camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,44 @@
...
+
+The team has agreed on an architecture: an abstract camera endpoint drives session lifecycle through events, while protocol-specific endpoints carry the signaling details. This keeps the client model clean — clients follow a single session flow regardless of whether the underlying camera uses WebRTC or OpenHome tunnels.
+
+## What Changes
+
+- **Abstract camera endpoint profile** (`profile: "camera"`): Define three resources on the camera endpoint — `startSession` [execute], `stopSession` [execute], and `sessionStatus` [emit events only]. `sessionStatus` is the "traffic controller" that directs clients through the session flow. Events carry metadata with `sessionId`, `technology`, and `nextAction` (full URI to the next resource the client should interact with).
```
rchowdcmcsa: > nextAction (full URI to the next resource the client should interact with).

Interact with how? For example, if nextAction points to peerSdp, a write resource, then we can surmise that the interaction is a write (and indeed it is). What about an event resource like tunnelUrl? Subscription? Or is it just simply informing the client, "hey, I'm sending the tunnel URL your way via event right after this"? Maybe this is obvious/self-evident but it's just unclear to me.

rchowdcmcsa (commented):
My bad if my comments overlap with anything that was said prior, I lost track of stuffs as I read through all this.

tleacmcsa:
We should probably consider (or heck, even implement?) support for Reolink cameras which have embraced local access and integrations for Home Assistant.  For reference: https://github.com/starkillerOG/reolink_aio 

Reolink maybe the same or very close to onvif (they seem to offer onvif interfaces).

We would end up with two very useful classes of camera integrations I think: remotely accessible (webrtc, relay server open home), and local only (onvif / reolink).  Would that be the same camera interface? clients would need to know what kind of access a camera driver provides.  Food for thought, but important to consider as we attempt to add some kind of flexible camera data model.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Goals / Non-Goals
+
+**Goals:**
+- Define an abstract camera session contract that is independent of any specific streaming protocol
+- Support WebRTC (Matter 1.5) and OpenHome cameras through the same client-facing session flow
+- Use `updateResource()` event metadata as the session coordination mechanism — no new infrastructure needed
+- Keep events as the source of truth for session state, enabling future parallel session support
+- Maintain full backward compatibility with existing OpenHome camera clients
+- Ensure the pattern extends naturally to future technologies (RTSP, HTTP, etc.)
```
tleacmcsa, line 19: i still think a goal would be to have a reusable webrtc profile.  It could be used for other non-camera use cases like ssh, binary upload/download, audio streaming, etc.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+Matter 1.5 introduces WebRTC-based cameras with a fundamentally different signaling pattern: SDP offer/answer exchange, ICE candidate negotiation, and a bidirectional transport flow. The team has agreed that rather than building another technology-specific camera model, Barton should define a protocol-agnostic session contract that both WebRTC and OpenHome cameras (and future technologies) can implement.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Define an abstract camera session contract that is independent of any specific streaming protocol
+- Support WebRTC (Matter 1.5) and OpenHome cameras through the same client-facing session flow
+- Use `updateResource()` event metadata as the session coordination mechanism — no new infrastructure needed
+- Keep events as the source of truth for session state, enabling future parallel session support
+- Maintain full backward compatibility with existing OpenHome camera clients
```
tleacmcsa: Another goal should probably be stated: support cameras that are local-only (no "session" negotiation)

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+```
+Camera device (deviceClass: "camera")
+├── ep/camera  (profile: "camera")         ← abstract session lifecycle
+│   ├── r/startSession    [execute]
+│   ├── r/stopSession     [execute]
+│   └── r/sessionStatus   [events only]    ← traffic controller
+│
+├── ep/webrtc  (profile: "webrtc")         ← only on WebRTC cameras
+│   ├── r/peerSdp              [write]
+│   ├── r/cameraSdp            [events]
```
tleacmcsa: the mention of camera in the webrtc profile seems unnecessary and may be inappropriate if this is used for non-camera use cases.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+#define CAMERA_SESSION_METADATA_NEXT_ACTION  "nextAction"
+```
+
+**Alternatives considered**:
+- GObject signals for session state: Would bypass the resource model entirely. Signals are appropriate for non-device-scoped events (system-level), but camera sessions are device-scoped and should flow through the resource/event system like all other device state.
+- New metadata enforcement mechanism: Would require changes across core, API, and all drivers. Over-engineering for MVP — the JSON convention is sufficient and extensible.
+
+### D3: Events are the source of truth, not resource values
+
+**Decision**: `sessionStatus` must emit events even when the underlying value doesn't change. The event (with metadata) is what drives client behavior. The stored resource value is secondary.
```
tleacmcsa: Is there a stored resource value?  There doesn't have to be.  It can be non-readable, but emit events on change (for the event + metadata).  Given that we need to support simultaneous sessions, im not sure reading this even makes sense unless it was some catalog of all active sessions.

update: oh, i see you mention some of this on the next line :-)

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+
+**Rationale**: In parallel session scenarios (future), multiple sessions may be in different states simultaneously. A single resource value can't represent this. Events with per-session metadata can. Even for single-session MVP, building on events-as-truth ensures the contract doesn't need to change when parallel sessions are added.
+
+`updateResource()` supports unconditional event emission for resources registered with `CACHING_POLICY_NEVER` and `RESOURCE_MODE_EMIT_EVENTS` — it skips value comparison entirely because non-cached resources have no previous value to compare against. Register `sessionStatus` with this combination and every `updateResource()` call will fire an event regardless of whether the value changed. No core changes are needed.
+
+**Session flow example (WebRTC)**:
+```
+Client                        Driver                    Camera
+  |                             |                          |
+  |── executeResource           |                          |
```
tleacmcsa, line 126: the client needs to know its session id, so it can know which sessionStatus events apply to it.  executeResource can return data and probably should return that before any events are emitted. otherwise the client should probably be required to provide a unique identifier as an input arg which is then used with the events to tie it together.  all subsequent requests from the same client about that same session should include the identifier.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+
+**Decision**: The `sessionStatus` resource emits events with `cJSON *metadata` carrying `sessionId`, `technology`, and `nextAction`. This reuses the existing `updateResource(deviceUuid, endpointId, "sessionStatus", stateValue, metadata)` mechanism.
+
+**Rationale**: `updateResource()` with metadata is already proven across the codebase — door lock uses it for lock source/userId, sensors for test fault flags. No new infrastructure is required. The metadata is "over the top" — a JSON convention between driver and client, not enforced by Barton's type system. This matches the team's explicit design decision.
+
+**Metadata schema** (JSON convention):
+```json
+{
+  "sessionId": "uuid-string",
+  "technology": "webrtc" | "openhome",
```
tleacmcsa: should this be "technology" or "protocol" ?

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,183 @@
...
+
+**Rationale**: A device's protocol capabilities are a function of its technology. Matter 1.5 cameras always use WebRTC. OpenHome cameras always use media tunnels. There's no runtime negotiation needed for MVP. The endpoint-existence pattern is forward-compatible — a future multi-protocol device could have both endpoints, and the driver (or a negotiation heuristic) could choose.
+
+**Alternatives considered**:
+- Protocol field on the device: Would require a new device-level property and introduce a code path distinct from the natural "what endpoints does this device have?" query.
+- Client-specified protocol: Clients shouldn't need to know which protocol a camera uses — that defeats protocol agnosticism.
+
+### D5: Backward compatibility with existing OpenHome camera profile
+
+**Decision**: The existing `CAMERA_PROFILE` resources (`portNumber`, `adminUserId`, `createMediaTunnel`, etc.) remain on the camera endpoint unchanged. The new session resources (`startSession`, `stopSession`, `sessionStatus`) are added alongside them. The `openhome` endpoint is new.
```
tleacmcsa: if the impact is not too great, we can consider breaking the existing resource model and fixing the only client (zilker).  I would rather not bring ugliness like "adminUserId" forward just for compatibility if we can easily fix it too.

openspec/changes/camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Non-goals
+
+- SBMD feasibility for Matter cameras needs evaluation but is out of scope for this proposal
+- Strongly typed metadata (enforced by Barton's type system) is a future improvement — metadata is "over the top" JSON convention for now
+- Other camera technologies (RTSP, plain HTTP) are not yet designed for, though the pattern should extend to them
+- Video rendering, decoding, or media pipeline — clients own that
+- Cloud relay or TURN server integration
+- Doorbell-specific device types (Video Doorbell 0x0143, Audio Doorbell 0x0141) — future work
+- Parallel session management design (future — the contract supports it via session IDs, but initial implementation is single-session)
```
tleacmcsa: we dont want to break the interface when we add parallel sessions, so we need to think that through if we can.

openspec/changes/camera-architecture-redesign/tasks.md
tleacmcsa: and some cameras/protocols which are direct only may not actually 'start' anything and instead just provide a URL.

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+
+## 1. Camera session contract and OpenHome adapter
+
+**Capabilities**: `camera-session-contract`, `openhome-session-adapter`
+
+Define the protocol-agnostic camera session contract and prove it end-to-end by adapting the existing OpenHome camera driver. This is the foundational work — it lands the architecture in code using existing hardware with no new external dependencies.
+
+Scope includes:
+- Session lifecycle resources (`startSession`, `stopSession`, `sessionStatus`) and metadata constants (`sessionId`, `technology`, `nextAction`, status values) in `commonDeviceDefs.h`
+- Coexistence with existing `CAMERA_PROFILE_*` definitions (backward compatibility)
```
tleacmcsa: backwards compatibility within reason.  We are in a unique position whereby we can fix clients at the same time if we need to make sensible changes to keep it clean.

tleacmcsa (changes requested):

openspec/changes/camera-architecture-redesign/tasks.md
kfundecmcsa: Agreed - reworded this to frame SBMD as the default and called out the composition problem as the key challenge for Task 2's explore phase. Details left to that task.

openspec/changes/camera-architecture-redesign/tasks.md
kfundecmcsa: copilot getting a bit over zealous, missed this one

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: done

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
kfundecmcsa: good point - done

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: agreed, done

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: this schema has been update/simplified so this `technology` field has been dropped, please see what you think

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: the schema has been updated/simplified, so that `technology` key is gone, so I think we are good. please take a look and see what you think

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Addressed - the updated D3 now explicitly states that `streamStatus` is non-readable (no stored value), events only.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
kfundecmcsa: Agreed - `executeResource("stream")` now returns the `streamId` synchronously. The client needs the ID before the first `streamStatus` event arrives so it can filter. All subsequent events include `streamId` in metadata for correlation. How we guarantee the ordering is TBD (see design.md Open Question #3 / Task 1 scope).

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
kfundecmcsa: Valid concerns. `stopSession` has been removed - driver self-manages lifecycle now. But your core question still applies: what happens when things end unexpectedly?

The intent is that the driver always emits a terminal streamStatus event (`"error"` or `"done"`) when a stream ends, regardless of reason. That way the client knows it's over.

The specifics of cleanup will get fleshed out in the follow-up implementation tasks.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Good question. Parallel streams are supported from the start now - each `stream` call returns a new `streamId` and the driver manages them independently. So calling `stream` while one is active just starts another one. Added clarification under D7.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Yeah, both paths could be exercised simultaneously. They're effectively independent - the new `stream` path calls `createMediaTunnel` internally and emits `streamStatus` events for tracking, while the old direct `createMediaTunnel` path just creates a tunnel without any stream tracking. They don't conflict, they just don't know about each other. If/when we deprecate the old resources this goes away.

Any concerns?

openspec/changes/camera-architecture-redesign/proposal.md
kfundecmcsa: My initial assumption at a single stream MVP was wrong. Removed all mentions of single sessions and the MVP will support parallel streams.

openspec/changes/camera-architecture-redesign/proposal.md
kfundecmcsa: Good call - this was a bit ambiguous before. In the updated design, `nextAction` always points to an execute resource - it just means "call this next." The client loop is the same regardless of protocol: get a `streamStatus` event, check for `nextAction`, execute it if present. Protocol-specific results (SDP answers, stream URLs, etc.) come back as events on the protocol endpoint's resource. No `nextAction` on the final `done` event means you're finished.

openspec/changes/camera-architecture-redesign/tasks.md
kfundecmcsa: Agreed. Rephrased to indicate clean break from these contracts is preferred, unless there's a compelling reason to keep them.

openspec/changes/camera-architecture-redesign/tasks.md
kfundecmcsa: Agreed on both counts. Going with `stream` for now (although that could change) as a single [execute] resource rather than separate `startSession`/`stopSession`. "start" was intentionally omitted from the name (not `startStream`) since, as Thomas pointed out, some cameras/protocols don't actually "start" anything.

Driver self-manages lifecycle (D7 in the updated design) - timeouts, cleanup, etc. are all internal.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Agreed - no reason to carry `adminUserId` and friends forward if we can clean it up. Updated D5 to make clean break the default. Legacy resources get removed unless there's a compelling reason to keep them. See design.md D5 and the Task 1 scope.

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+Implement the Matter 1.5 WebRTC camera driver and signaling flow on top of the stream contract from Task 1.
+
+Scope includes:
+- `webrtc` endpoint profile with `offerSdp`, `remoteSdp`, `offerIceCandidates`, `remoteIceCandidates` resources and `WEBRTC_PROFILE_*` constants
+- Matter camera driver via SBMD — the key architectural challenge is SBMD composition: how to reuse the shared camera stream manager (stream lifecycle, status events, cleanup) from within an SBMD driver. Options to explore include SBMD class dependencies and exposing the stream manager via JS bindings in the SBMD context.
+- Cluster server hosting pattern (`WebRtcTransportRequestorServer`) — first server-side cluster in Barton's Matter subsystem
+- SDP offer/answer exchange and ICE candidate negotiation
+- WebRTC media stack integration (libdatachannel or alternative)
+- Integration test infrastructure: matter.js virtual camera device, Python test fixtures (`MatterCamera`) following the existing `MatterLight`/`MatterDoorLock` pattern
+
```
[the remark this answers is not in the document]
kfundecmcsa: We are purposefully experimenting with this workflow wherein the tasks created by openspec are themselves setup for future work. This may break with openspec convention, but this method seemed like a practical way to divy up complex work amongst the team.

openspec/changes/camera-architecture-redesign/design.md
rchowdcmcsa: Sounds alright to me.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Capabilities
+
+### New Capabilities
+- `camera-stream-contract`: The abstract camera endpoint profile with `stream` and `streamStatus` resources, event metadata schema, and stream lifecycle flow. This is the protocol-agnostic core of the design.
+- `webrtc-endpoint-profile`: The WebRTC protocol-specific endpoint profile (`offerSdp`, `remoteSdp`, `offerIceCandidates`, `remoteIceCandidates`) and the Matter 1.5 WebRTC signaling flow that drives it.
+- `openhome-stream-adapter`: Adaptation of the existing OpenHome camera driver to emit stream events through the new abstract camera endpoint contract, with the `openhome` profile providing `streamUrl`.
+
+### Modified Capabilities
+_(none — no existing spec-level requirements are changing)_
```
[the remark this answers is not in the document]
kfundecmcsa: This commit does not introduce changes yet, it is setting up changes for the future. There will be both modified and new capabilities, but those will be explicitly defined in the upcoming changes.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,212 @@
...
+**Rationale**: `updateResource()` with metadata is already proven across the codebase — door lock uses it for lock source/userId, sensors for test fault flags. No new infrastructure is required. The metadata is "over the top" — a JSON convention between driver and client, not enforced by Barton's type system. This matches the team's explicit design decision.
+
+**Stream status values**: `setup`, `done`, `error` — intentionally minimal to work across all camera types regardless of protocol complexity.
+
+**Metadata schema** (JSON convention):
+```json
+{
+  "streamId": "<uint64>",
+  "nextAction": "/deviceId/ep/webrtc/r/offerSdp"
+}
```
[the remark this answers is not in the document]
kfundecmcsa: done

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,212 @@
+## Context
+
+BartonCore provides protocol-agnostic IoT device management through a URI-based resource model (`/deviceId/ep/N/r/resourceName`). Devices have endpoints with profiles (e.g., `doorLock`, `sensor`, `light`) that define mandatory and optional resources. Drivers communicate state changes via `updateResource()`, which can carry optional `cJSON *metadata` — a pattern already used by door lock (source, userId), sensor (test flag), and link quality drivers.
```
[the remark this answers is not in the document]
kfundecmcsa: done

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,212 @@
...
+
+**Stream status values**: `setup`, `done`, `error` — intentionally minimal to work across all camera types regardless of protocol complexity.
+
+**Metadata schema** (JSON convention):
+```json
+{
+  "streamId": "<uint64>",
+  "nextAction": "/deviceId/ep/webrtc/r/offerSdp"
+}
+```
```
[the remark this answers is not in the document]
kfundecmcsa: Fair - I adjusted to only use absolute paths

openspec/changes/camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,43 @@
...
+- **Stream coordination via `updateResource()` event metadata**: Reuse the existing `cJSON *metadata` mechanism already used by door lock, sensor, and link quality drivers. The metadata carries stream state as a JSON convention between driver and client — not enforced by Barton's type system. Events are the source of truth for stream state, not resource values. Protocol-specific data (SDP answers, ICE candidates, URLs) is emitted on protocol endpoint resources, not on `streamStatus`.
+- **Driver-determined protocol selection**: The driver inspects which protocol endpoints exist on the device to determine the signaling flow. For MVP, this is hardcoded per driver (Matter camera driver → WebRTC, OpenHome camera driver → openhome).
+- **Existing camera profile constants**: The existing `CAMERA_PROFILE_*` definitions in `commonDeviceDefs.h` are retained where practical. Since we own the clients, a clean break is acceptable if it simplifies the design — backward compatibility is preferred but not a hard constraint.
+
```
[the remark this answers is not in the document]
kfundecmcsa: Fair point - I adjusted the wording to be more concrete and consistent across the proposal

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+**Capabilities**: `webrtc-endpoint-profile`
+**Depends on**: Task 1 (camera stream contract)
+
+Implement the Matter 1.5 WebRTC camera driver and signaling flow on top of the stream contract from Task 1.
+
+Scope includes:
+- `webrtc` endpoint profile with `offerSdp`, `remoteSdp`, `offerIceCandidates`, `remoteIceCandidates` resources and `WEBRTC_PROFILE_*` constants
+- Matter camera driver via SBMD — the key architectural challenge is SBMD composition: how to reuse the shared camera stream manager (stream lifecycle, status events, cleanup) from within an SBMD driver. Options to explore include SBMD class dependencies and exposing the stream manager via JS bindings in the SBMD context.
+- Cluster server hosting pattern (`WebRtcTransportRequestorServer`) — first server-side cluster in Barton's Matter subsystem
```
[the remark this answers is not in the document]
kfundecmcsa: done

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,212 @@
...
+**Alternatives considered**:
+- Protocol field on the device: Would require a new device-level property and introduce a code path distinct from the natural "what endpoints does this device have?" query.
+- Client-specified protocol: Clients shouldn't need to know which protocol a camera uses — that defeats protocol agnosticism.
+
+### D5: Backward compatibility with existing OpenHome camera profile
+
+**Decision**: The new stream resources (`stream`, `streamStatus`) are added to the camera endpoint. The `openhome` endpoint is new. Legacy resources (`createMediaTunnel`, `adminUserId`, etc.) are expected to be removed unless there's a compelling reason to keep them — we own the clients and would rather not carry forward ugly naming just for compatibility.
+
+**Rationale**: The team owns all camera clients and can update them alongside the service. A clean break is the default — backward compatibility is only preserved if the cost of removing it is surprisingly high.
+
```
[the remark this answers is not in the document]
kfundecmcsa: Fair point - renamed this to `Legacy camera profile removal`

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,212 @@
...
+## Goals / Non-Goals
+
+**Goals:**
+- Define an abstract camera stream contract that is independent of any specific streaming protocol
+- Support WebRTC (Matter 1.5) and OpenHome cameras through the same client-facing stream flow
+- Support local-only cameras that require no negotiation (e.g., ONVIF, Reolink) — the contract handles both multi-step signaling and immediate URL delivery
+- Design the WebRTC endpoint profile to be reusable beyond cameras (e.g., ssh, binary transfer, audio streaming)
+- Use `updateResource()` event metadata as the stream coordination mechanism — no new infrastructure needed
+- Keep events as the source of truth for stream state, with per-stream correlation via `streamId`
+- Maintain full backward compatibility with existing OpenHome camera clients
```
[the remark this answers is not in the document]
kfundecmcsa: fair - done

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
cleithner-comcast: The ordering should be natural since any given client command on an endpoint profile would execute instructions within the same thread as each other. That is, there's no concurrency in "HandleExecuteOfferSDP" or whatever. In some extreme case that I can't think of, an endpoint profile could implement a priority queue (timestamp) for events before emitting, or simply include nondeterministic event ordering in its contract.

The only sticky point to me is that you need to emit the first streamStatus event AFTER the client has received the streamID, which is completely opaque to the driver. Emitting it after some delay (via glib main loop scheduleing) is probably the easiest way but it is somewhat risky. There's no guarantee the client will receive the execute resource return value before they get that event.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+`updateResource()` supports unconditional event emission for resources registered with `CACHING_POLICY_NEVER` and `RESOURCE_MODE_EMIT_EVENTS` — it skips value comparison entirely because non-cached resources have no previous value to compare against. Register `streamStatus` with this combination and every `updateResource()` call will fire an event regardless of whether the value changed. No core changes are needed.
+
+**Stream flow example (WebRTC)**:
+```
+Client                        Driver                    Camera
+  |                             |                          |
+  |── executeResource           |                          |
+  |   (stream, {})  ──────►     |                          |
+  |   ◄── returns streamId      |                          |
+  |                             |── (allocate stream) ───► |
```
cleithner-comcast: This is not a real step. It may be for some technologies but generally you start with discovery and then some kind of protocol messaging after. It will often be correct that once a streamId is created (and known to the client) the next step is barton emitting the stramStatus event without interacting with the camera just yet until the client executes an endpoint profile resource.

cleithner-comcast (dismissed):

openspec/changes/camera-architecture-redesign/design.md
rchowdcmcsa: Looks good to me!

rchowdcmcsa (dismissed):

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+- Strongly typed metadata enforcement (future improvement)
+- SBMD feasibility evaluation for Matter cameras (separate investigation)
+- Video rendering, decoding, or media pipeline integration
+- WebRTC media stack selection (libdatachannel vs. alternatives — deferred to webrtc-endpoint-profile design)
+
+## Decisions
+
+### D1: Two-layer endpoint architecture — abstract + protocol-specific
+
+**Decision**: Every camera device has an abstract **camera endpoint** (`profile: "camera"`) that owns stream lifecycle, plus one or more **protocol-specific endpoints** that carry the signaling details for a particular technology.
```
tleacmcsa: Could the main camera endpoint handle local-only cameras due to the simplicity (return a URL)?

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+
+### D2: Stream coordination via `updateResource()` event metadata
+
+**Decision**: The `streamStatus` resource emits events with `cJSON *metadata` carrying `streamId` and `nextAction`. This reuses the existing `updateResource(deviceUuid, endpointId, "streamStatus", stateValue, metadata)` mechanism. The `stream` execute resource returns the `streamId` synchronously so the client has a correlation ID before any events fire.
+
+**Rationale**: `updateResource()` with metadata is already proven across the codebase — door lock uses it for lock source/userId, sensors for test fault flags. No new infrastructure is required. The metadata is "over the top" — a JSON convention between driver and client, not enforced by Barton's type system. This matches the team's explicit design decision.
+
+**Stream status values**: `setup`, `done`, `error` — intentionally minimal to work across all camera types regardless of protocol complexity.
+
+**Metadata schema** (JSON convention — `streamId` is a driver-generated counter, unique per device):
```
tleacmcsa: counter is an implementation detail.  We may likely find that there are issues with that approach and need to use UUID or something.  I would just call it a "unique id"

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+  "streamId": 1,
+  "nextAction": "/deviceId/ep/webrtc/r/offerSdp"
+}
+```
+
+On error:
+```json
+{
+  "streamId": 1,
+  "error": "<description>"
```
tleacmcsa, line 91: does an "error" element mean the stream is dead?  If there is an error on the initial execute stream resource does it return a unique id at all?  Maybe too low level for now, but something to consider in the design.

tleacmcsa: also, how does a client know when the stream is dead (the unique id is closed or invalidated)?  A camera could reboot or report some error condition out of band from the client's direct connection.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+- GObject signals for stream state: Would bypass the resource model entirely. Signals are appropriate for non-device-scoped events (system-level), but camera streams are device-scoped and should flow through the resource/event system like all other device state.
+- New metadata enforcement mechanism: Would require changes across core, API, and all drivers. Over-engineering for MVP — the JSON convention is sufficient and extensible.
+- Granular status values (`negotiating`, `ice-exchange`, `connected`): Too protocol-specific — `negotiating` doesn't apply to OpenHome or direct cameras. `setup`/`done`/`error` works universally.
+
+### D3: Events are the source of truth, not resource values
+
+**Decision**: `streamStatus` is non-readable (no stored value) and emits events unconditionally. The event (with metadata) is what drives client behavior.
+
+**Rationale**: In parallel stream scenarios (future), multiple streams may be in different states simultaneously. A single resource value can't represent this. Events with per-stream metadata can. Even for single-stream MVP, building on events-as-truth ensures the contract doesn't need to change when parallel streams are added.
```
tleacmcsa: maybe we should drop terminology like "future".  We are designing the interface here and we want parallel stream support in that from the beginning.  Also, i think we realized that it isnt any harder to do it.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+  |── executeResource           |                          |
+  |   (getStreamUrl) ──────►    |── createMediaTunnel() ─► |
+  |                             |                          |
+  | ◄── streamUrl event ──────  |   ← protocol endpoint    |
+  |   value: "rtsp://..."       |                          |
+  |                             |                          |
+  | ◄── streamStatus event ───  |                          |
+  |   value: "done"             |                          |
+  |   meta: {streamId}          |                          |
+```
```
tleacmcsa, line 188: Can you add a local-only example where a simple URL is returned?

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+  |                             |                          |
+  | ◄── streamUrl event ──────  |   ← protocol endpoint    |
+  |   value: "rtsp://..."       |                          |
+  |                             |                          |
+  | ◄── streamStatus event ───  |                          |
+  |   value: "done"             |                          |
+  |   meta: {streamId}          |                          |
+```
+
+### D4: Driver determines protocol by inspecting endpoint existence
```
tleacmcsa: I don't follow this.  There should be a driver per camera protocol, not one big camera driver that keeps getting updated as new protocols emerge.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Rationale**: A device's protocol capabilities are a function of its technology. Matter 1.5 cameras always use WebRTC. OpenHome cameras always use media tunnels. There's no runtime negotiation needed for MVP. The endpoint-existence pattern is forward-compatible — a future multi-protocol device could have both endpoints, and the driver (or a negotiation heuristic) could choose.
+
+**Alternatives considered**:
+- Protocol field on the device: Would require a new device-level property and introduce a code path distinct from the natural "what endpoints does this device have?" query.
+- Client-specified protocol: Clients shouldn't need to know which protocol a camera uses — that defeats protocol agnosticism.
+
+### D5: Legacy camera profile removal
+
+**Decision**: The new stream resources (`stream`, `streamStatus`) are added to the camera endpoint. The `openhome` endpoint is new. Legacy resources (`createMediaTunnel`, `adminUserId`, etc.) are expected to be removed unless there's a compelling reason to keep them — we own the clients and would rather not carry forward ugly naming just for compatibility.
```
tleacmcsa: the local-only support will need credentials to log into the camera somewhere, hmm.  maybe we figure that out later and for now assume its in the URL as query string args (yuck).

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Rationale**: The team owns all camera clients and can update them alongside the service. A clean break is the default — backward compatibility is only preserved if the cost of removing it is surprisingly high.
+
+**Thread safety**: `updateResource()` with metadata runs on the GLib main loop. Matter subsystem callbacks marshal to the GLib main loop via `g_main_context_invoke` before calling `updateResource()`. This follows the established pattern used by Matter subscription callbacks. No additional threading concerns.
+
+### D6: Matter 1.5 cameras use WebRTC transport
+
+**Decision**: Matter 1.5 introduces camera streaming via WebRTC signaling clusters — `WebRTCTransportProvider` on the commissioner (Barton) and `WebRTCTransportRequestor` on the camera. Barton's Matter camera driver uses this transport, mapping the WebRTC signaling flow onto the stream contract defined in D1–D3.
```
tleacmcsa: in this context, Barton is the "controller" not "commissioner".

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Rationale**: WebRTC is the only streaming transport defined in the Matter 1.5 camera specification. There is no alternative to evaluate — this is the protocol.
+
+**Architectural implication**: This requires Barton to host a cluster server for the first time — receiving callbacks from the camera device rather than only sending commands. The detailed signaling flow, cluster server hosting pattern, media stack selection (libdatachannel or alternative), and thread marshalling design are scoped to the `webrtc-endpoint-profile` task.
+
+### D7: Driver self-manages stream lifecycle
+
+**Decision**: There is no `stopStream` resource. The driver owns stream lifecycle — it handles timeouts, device disconnects, and cleanup internally. When a stream ends (for any reason), the driver emits `streamStatus: "done"` (or `"error"`).
```
tleacmcsa: Might mention that local-only cameras may not manage the streams and likely just pass a URL to the client.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Risks / Trade-offs
+
+- **[Metadata is untyped]** → Mitigation: Well-documented JSON schema, string constants in `commonDeviceDefs.h`, and integration tests that verify metadata content. Strongly typed metadata is a future capability.
+- **[SBMD composition for Matter cameras]** → Mitigation: The Matter camera driver should be SBMD, but SBMD doesn't currently support class dependencies. The key challenge is making the shared camera stream manager available in the SBMD JS context. This is scoped to the webrtc-endpoint-profile task.
+- **[OpenHome adapter complexity]** → Mitigation: The OpenHome driver already implements `createMediaTunnel`. The adapter wraps this in the stream contract — `stream` execute calls `createMediaTunnel` internally and emits `streamStatus` events. Incremental, low-risk change.
+
+## Open Questions
+
+1. **What happens to in-flight streams on driver restart?** The driver should emit `streamStatus: "error"` for all active streams on shutdown. Detail in the parallel stream management task.
```
tleacmcsa: might add "for streams managed by the driver" to be clear that local-only cameras dont need to do that.

tleacmcsa (changes requested):
I reviewed design.md before proposal.md (oops), so my comments are there, but they likely apply to the proposal.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
kfundecmcsa: good call, done

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Good catch, removed this and all other references to future parallel work

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: good call, done

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Good call - that was misleading. Removed this bit to line up with what `createSession` actually does.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Agreed - that was the intent, but not all that clear. Rewrote D4 to make it explicit: one driver per protocol, each creates its own endpoints.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Agreed - credentials for local cameras are an open question. Added it explicitly to D8. Options range from URL-embedded creds to separate resources on the `direct` endpoint to config-time setup. TBD and likely vendor-dependent.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Good catch - fixed

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: Covered - added D8 to provide more detail on local/direct cameras.

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: good call - done

openspec/changes/camera-architecture-redesign/design.md
kfundecmcsa: We discussed this and went with a separate `ep/direct` endpoint. Keeps the pattern consistent - every protocol gets its own endpoint, client flow is identical regardless of technology.

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
kfundecmcsa: Good callout.

As of now, `sessionStatus: "error"` means the session is dead - the driver emits this for external failures (camera reboot, disconnect), which is how the client knows out-of-band. If `createSession` itself fails, the execute returns `false` with no sessionId.

That said, the finer details (auto-invalidation vs. still requiring `destroySession` after error, what a client should do on reconnect, timeout strategies) are still open - added an "Error behavior" section to D9 to acknowledge these are implementation-time decisions.

tleacmcsa (dismissed):

cleithner-comcast (changes requested):
This seems to still be missing explicit actions on the abstract endpoint. Although you have createSession, there's no "stream" or "takePicture" resource on the camera endpoint. I think those would be the entry points for actual actions (so that nextAction makes sense per each action). createSession just gives you the identifier to start one or each of those.

openspec/changes/camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Risks / Trade-offs
+
+- **[Metadata is untyped]** → Mitigation: Well-documented JSON schema, string constants in `commonDeviceDefs.h`, and integration tests that verify metadata content. Strongly typed metadata is a future capability.
+- **[SBMD composition for Matter cameras]** → Mitigation: The Matter camera driver should be SBMD, but SBMD doesn't currently support class dependencies. The key challenge is making the shared camera session manager available in the SBMD JS context. This is scoped to the webrtc-endpoint-profile task.
+- **[OpenHome adapter complexity]** → Mitigation: The OpenHome driver already implements `createMediaTunnel`. The adapter wraps this in the session contract — `createSession` execute calls `createMediaTunnel` internally and emits `sessionStatus` events. The URL is delivered via `mediaUrl` on the openhome endpoint. Incremental, low-risk change.
```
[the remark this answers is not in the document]
kfundecmcsa: good catch, fixed

openspec/changes/archive/2026-05-04-camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,41 @@
...
+- **Existing `camera` profile**: Extended, not replaced. New session resources coexist with existing OpenHome camera resources.
+- **Event metadata contract**: New JSON keys (`sessionId`, `protocol`, `nextAction`) added as string constants in `commonDeviceDefs.h`. These are conventions, not type-system enforced.
+- **`sessionStatus` is non-readable, events only**: Registered with `CACHING_POLICY_NEVER` and `RESOURCE_MODE_EMIT_EVENTS`, which skips value comparison entirely. Every `updateResource()` call fires an event.
+- **CMake flags**: `BCORE_MATTER` for Matter camera support. WebRTC media dependencies TBD (to be designed in the webrtc-endpoint-profile capability).
+- **Backward compatibility**: Not a goal. We own the clients and would rather not carry forward ugly legacy naming. Clean break is the default — old resources are removed unless there's a compelling reason to keep them.
```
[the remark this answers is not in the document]
kfundecmcsa: I think the goal has been made clear enough in the proposal - attempt clean break unless it gets too messy

openspec/changes/camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
...
+
+Scope includes:
+- Session lifecycle resources (`createSession` [execute, returns sessionId], `destroySession` [execute], `sessionStatus` [events only, non-readable]) and metadata constants (`sessionId`, `protocol`, `nextAction`, status values `setup`/`done`/`error`) in `commonDeviceDefs.h`
+- Event ordering guarantee: the client must receive the `sessionId` from the `createSession` execute response before the first `sessionStatus` event fires. Mechanism TBD (scheduled delay, deferred emit, etc.)
+- Clean up existing `CAMERA_PROFILE_*` definitions — legacy resources are removed unless there's a compelling reason to keep them (we own the clients)
+- `openhome` endpoint profile with `getMediaUrl` [execute] and `mediaUrl` [events] resources and `OPENHOME_PROFILE_*` constants
+- OpenHome camera driver adaptation: `createSession` execute internally calls `createMediaTunnel`, emits `sessionStatus` events with metadata. Existing `createMediaTunnel` path may be retained or removed depending on whether a clean break simplifies the implementation.
+- Session ID lifecycle (client calls `destroySession` to clean up; driver cleans up on device removal/restart)
```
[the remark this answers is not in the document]
kfundecmcsa: fair - done

kfundecmcsa:
> This seems to still be missing explicit actions on the abstract endpoint. Although you have createSession, there's no "stream" or "takePicture" resource on the camera endpoint. I think those would be the entry points for actual actions (so that nextAction makes sense per each action). createSession just gives you the identifier to start one or each of those.

Great catch - this has been updated to include these resources in the profile/flow examples. Please take a look when you can.

cleithner-comcast (dismissed):

cleithner-comcast (approved):

tleacmcsa (approved):

openspec/changes/archive/2026-05-04-camera-architecture-redesign/tasks.md
```diff
@@ -0,0 +1,31 @@
+_These tasks represent follow-on implementation work. Each is expected to go through its own explore → propose → apply cycle._
+
+## 1. Camera session contract and OpenHome adapter
+
+**Capabilities**: `camera-session-contract`, `openhome-stream-adapter`
+
+Define the protocol-agnostic camera session contract and prove it end-to-end by adapting the existing OpenHome camera driver. This is the foundational work — it lands the architecture in code using existing hardware with no new external dependencies.
+
+Scope includes:
+- Session lifecycle resources (`createSession` [execute, returns sessionId], `destroySession` [execute], `sessionStatus` [events only, non-readable]) and metadata constants (`sessionId`, `protocol`, `nextAction`, status values `setup`/`done`/`error`) in `commonDeviceDefs.h`
```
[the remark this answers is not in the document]
kfundecmcsa: Team has approved  as is, not implementing this change

openspec/changes/archive/2026-05-04-camera-architecture-redesign/design.md
```diff
@@ -0,0 +1,283 @@
...
+
+**Endpoint structure per device class**:
+```
+Camera device (deviceClass: "camera")
+├── ep/camera  (profile: "camera")         ← abstract session lifecycle
+│   ├── r/createSession   [execute]        ← returns sessionId
+│   ├── r/stream          [execute]        ← start streaming (takes sessionId)
+│   ├── r/takePicture     [execute]        ← take snapshot (takes sessionId)
+│   ├── r/destroySession  [execute]        ← client cleanup
+│   └── r/sessionStatus   [events only]    ← traffic controller
```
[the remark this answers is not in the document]
kfundecmcsa: Team has approved  as is, not implementing this change

openspec/changes/archive/2026-05-04-camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,41 @@
...
+- **WebRTC endpoint profile** (`profile: "webrtc"`): Protocol-specific endpoint with resources for WebRTC signaling — `offerSdp` [execute], `remoteSdp` [events], `offerIceCandidates` [execute], `remoteIceCandidates` [events]. This endpoint exists only on devices that use WebRTC transport (Matter 1.5 cameras).
+- **OpenHome endpoint profile** (`profile: "openhome"`): Protocol-specific endpoint with `getMediaUrl` [execute] and `mediaUrl` [events]. This endpoint exists only on devices that use OpenHome media tunnels.
+- **Direct endpoint profile** (`profile: "direct"`): Protocol-specific endpoint for local cameras (ONVIF, Reolink, plain RTSP/HTTP) with `getMediaUrl` [execute] and `mediaUrl` [events]. Same resource pattern as OpenHome — the URL is simply available immediately without async negotiation.
```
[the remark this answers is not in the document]
kfundecmcsa: Team has approved  as is, not implementing this change

openspec/changes/archive/2026-05-04-camera-architecture-redesign/proposal.md
```diff
@@ -0,0 +1,41 @@
...
+
+## What Changes
+
+- **Abstract camera endpoint profile** (`profile: "camera"`): Define three resources on the camera endpoint — `createSession` [execute], `destroySession` [execute], and `sessionStatus` [emit events only, non-readable]. `createSession` returns a `sessionId` synchronously so the client has a correlation ID before events fire. `sessionStatus` is the "traffic controller" that emits lifecycle state transitions (`setup`, `done`, `error`) with metadata including `sessionId`, `protocol` (the technology in use), and `nextAction` (the next resource the client should interact with). Clients are required to call `destroySession` when done.
+- **WebRTC endpoint profile** (`profile: "webrtc"`): Protocol-specific endpoint with resources for WebRTC signaling — `offerSdp` [execute], `remoteSdp` [events], `offerIceCandidates` [execute], `remoteIceCandidates` [events]. This endpoint exists only on devices that use WebRTC transport (Matter 1.5 cameras).
```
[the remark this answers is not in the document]
kfundecmcsa: Team has approved  as is, not implementing this change

PR #200, opened 2026-04-07
ci: improve release workflow and regenerate full CHANGELOG

cleithner-comcast:
- Add branch guard (github.ref == refs/heads/main) to release job
- Convert lightweight tags to annotated after cocogitto bump
- Push version commit to main alongside tag
- Write release body to temp file instead of overwriting CHANGELOG.md
- Add branch_whitelist to cog.toml for defense in depth
- Regenerate CHANGELOG.md with full history from 0.9.3 through 4.0.0

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,6 +35,7 @@ on:
 
 jobs:
   release:
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yaml
```diff
@@ -49,16 +50,24 @@ jobs:
         with:
           release: true
 
-      - name: Push tag via Git
-        run: git push origin ${{ steps.release.outputs.version }}
+      - name: Convert to annotated tag
+        run: |
+          VERSION="${{ steps.release.outputs.version }}"
+          git tag -d "$VERSION"
```
tleacmcsa, line 56: why dis?

tleacmcsa (approved):

.github/workflows/release.yaml
```diff
@@ -49,16 +50,24 @@ jobs:
...
+          git tag -a "$VERSION" -m "$VERSION" HEAD
```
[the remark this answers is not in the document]
cleithner-comcast: Yeah I changed it to that cause I wanted it to be.

.github/workflows/release.yaml
cleithner-comcast: There's no configurable way to make cocogitto's tag annotated. So this just deletes the lightweight tag and remakes it as an annotated tag before pushing it.

kfundecmcsa: I think that reasoning probably warrants a comment, good to know for the future if this needs to be updated or changed.

kfundecmcsa (approved):

PR #201, opened 2026-04-08
build(docker): Add remote radio support in docker

Lakshmi97-velampati:
Add an optional otbr-radio Docker container that connects a Silicon Labs BRD2703 xG24 Explorer Kit to the Barton devcontainer via D-Bus, enabling Thread integration testing with real hardware.

- Dockerfile.otbr-radio: builds image with cpcd and otbr-agent compiled for CPC transport; uses the URL spinel+cpc://cpcd_0?iid=1&iid-list=0 to handle both unicast and broadcast Spinel frames
- compose.otbr-radio.yaml: compose overlay that shares a D-Bus socket volume with the barton service and maps RADIO_DEVICE at the same path on both the host and container sides
- otbr-radio-entrypoint.sh: writes a fresh cpcd config on every start (uart_device_file, uart_hardflow: true), waits for "Daemon startup was successful" before launching otbr-agent, and auto-detects BACKBONE_IF from the host default route at runtime
- setupDockerEnv.sh: auto-detects the backbone interface via ip route and writes it to docker/.env;

Refs: BARTON-359


docs/THREAD_BORDER_ROUTER_SUPPORT.md
```diff
@@ -1,3 +1,272 @@
...
+sudo apt install linux-tools-generic hwdata
+
+# Load the kernel module
+sudo modprobe vhci-hcd
+
+# List the available devices via the tunnel and confirm the BRD2703 is visible
+usbip list -r 127.0.0.1
+
+# Attach the remote device (replace the bus ID as appropriate)
+sudo usbip attach -r 127.0.0.1 -b <bus-id>
```
kfundecmcsa: for a lot of us on our remote dev machine, we don't have access to `sudo`. As a result, I wonder how this will affect this workflow. I don't have a solution necessarily, just pointing it out.

docs/THREAD_BORDER_ROUTER_SUPPORT.md
```diff
@@ -1,3 +1,272 @@
...
+
+### Devcontainer (VS Code)
+
+Add `compose.otbr-radio.yaml` to the `dockerComposeFile` array in
+`.devcontainer/devcontainer.json`:
+
+```jsonc
+"dockerComposeFile": [
+    "../docker/compose.devcontainer.yaml",
+    "../docker/compose.otbr-radio.yaml"   // add this line
```
kfundecmcsa: Editing the devcontainer.json file in order to get the thread support is a bummer. This file is checked in, so any changes here will be tracked. I wonder if there's a way to optionally rebuild the devcontainer with thread support that doesn't write to a checked in file? Not sure, worth investigating I think.

I do recall that it's tricky to do "stuff" before the devcontainer.json is loaded up, at the time, I think `initializeCommand` was the only method. Maybe we could hack onto that somehow.

An even better approach I think is we could start following Weston's example in https://github.com/rdk-gdcs/vsc-rdkb. He cleverly added an overlay configuration that dynamically builds the devcontainer.json file before the container is launched (see https://github.com/rdk-gdcs/vsc-rdkb/blob/master/.devcontainer/bin/workspace-init.sh). This would be a pretty big change to our build system, so might be out of scope for this effort, but this would resolve this issue, and any other potential devcontainer.json issues down the road.

kfundecmcsa (changes requested):
I have a couple open questions, but requesting changes for the title. This effort should fall into the `build` commit type, not `feat`

docker/otbr-radio-entrypoint.sh
```diff
@@ -0,0 +1,162 @@
...
+if [ -f /var/run/dbus/pid ]; then
+    dbusPid=$(cat /var/run/dbus/pid 2>/dev/null || true)
+    if [ -n "${dbusPid}" ] && kill -0 "${dbusPid}" 2>/dev/null; then
+        echo "[otbr-radio] D-Bus already running (PID ${dbusPid}), skipping start."
+    else
+        echo "[otbr-radio] Removing stale D-Bus pid/socket files..."
+        rm -f /var/run/dbus/pid /var/run/dbus/system_bus_socket
+        dbus-daemon --system --fork
+    fi
+else
```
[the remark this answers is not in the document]
Lakshmi97-velampati: Done

docker/otbr-radio-entrypoint.sh
```diff
@@ -0,0 +1,162 @@
...
+instance_name: cpcd_0
+bus_type: UART
+uart_device_file: ${RADIO_DEVICE}
+uart_device_baud: 115200
+uart_hardflow: true
+EOF
+
+echo "[otbr-radio] Starting cpcd (instance: cpcd_0, device: ${RADIO_DEVICE})..."
+CPCD_LOG="/tmp/cpcd.log"
+cpcd --conf "${CPCD_CONF}" 2>&1 | tee "${CPCD_LOG}" &
```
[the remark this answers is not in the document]
Lakshmi97-velampati: Done

docs/THREAD_BORDER_ROUTER_SUPPORT.md
Lakshmi97-velampati: @tleacmcsa has given limited access for usbip usage so these commands won't prompt for password. Able to use those.

Lakshmi97-velampati:
> I have a couple open questions, but requesting changes for the title. This effort should fall into the `build` commit type, not `feat`

Updated

docs/THREAD_BORDER_ROUTER_SUPPORT.md
cleithner-comcast: Meh, I'm not over the moon about the vsc-rdkb system. I don't like that it's extra tasks run before I can use the workspace. I wouldn't mind just composing otbr-radio by default if the service is harmless. I did this with observability services and it was fine.

docker/compose.otbr-radio.yaml
```diff
@@ -0,0 +1,114 @@
...
+        # privileged is required for:
+        #   - creating the wpan0 network interface
+        #   - configuring iptables rules for Thread border routing
+        #   - accessing the USB serial device via /dev/ttyACM*
+        privileged: true
+        # Host networking is required so that:
+        #   - the wpan0 Thread interface is visible on the host network stack
+        #   - mDNS/Avahi advertisements reach the LAN
+        #   - NAT64 routing functions correctly
+        network_mode: host
```
cleithner-comcast: I'm skeptical of this if USB/IP devices are properly forwarded to the container. Why would the custom docker network not work?

docker/compose.otbr-radio.yaml
```diff
@@ -0,0 +1,114 @@
...
+        volumes:
+            # Share the D-Bus socket directory with the barton/devcontainer
+            # service so Barton code can talk to otbr-agent over D-Bus.
+            - dbus-socket:/var/run/dbus
+        devices:
+            # Pass through the Silabs USB radio to the container at the same
+            # path as the host so RADIO_DEVICE matches inside the container.
+            - "${RADIO_DEVICE:-/dev/ttyACM0}:${RADIO_DEVICE:-/dev/ttyACM0}"
```
cleithner-comcast: How does this function if RADIO_DEVICE isn't populated or a user doesn't have one plugged in yet?

docker/compose.otbr-radio.yaml
```diff
@@ -0,0 +1,114 @@
...
+            - RADIO_DEVICE=${RADIO_DEVICE:-/dev/ttyACM0}
+            - BACKBONE_IF=${BACKBONE_IF:-}
+        # Restart automatically if cpcd/otbr-agent crash, unless the container
+        # was explicitly stopped.
+        restart: unless-stopped
+
+    # Extend the barton service to mount the shared D-Bus socket volume.
+    # This grants Barton access to the otbr-agent D-Bus API running in the
+    # otbr-radio container without any code changes.
+    # In devcontainer mode the devcontainer service extends barton, so it
```
cleithner-comcast: This isn't exactly true, we removed the extends clause. Devcontainer is now a layered approach

docker/otbr-radio-entrypoint.sh
```diff
@@ -0,0 +1,168 @@
...
+    echo "[otbr-radio] Auto-detected backbone interface: ${BACKBONE_IF}"
+fi
+
+###############################################################################
+# 1. Start D-Bus system bus
+#
+# The /var/run/dbus directory is mounted from the persistent named volume
+# 'dbus-socket'. If the container previously crashed or was restarted, the
+# pid file from the previous run may still be present, causing dbus-daemon to
+# refuse to start. Clean it up if the recorded PID is no longer alive.
```
cleithner-comcast: I would expect the container to not manage the system dbus socket since that's owned and maintained by the host. The commands below seem suspect. I don't think this container should ever be writing to the volume except to the socket for messaging.

docker/otbr-radio-entrypoint.sh
```diff
@@ -0,0 +1,168 @@
...
+    dbus-daemon --system --fork
+fi
+echo "[otbr-radio] D-Bus started."
+
+###############################################################################
+# 2. Validate the USB radio device
+###############################################################################
+if [ ! -e "${RADIO_DEVICE}" ]; then
+    echo "[otbr-radio] ERROR: USB radio device '${RADIO_DEVICE}' not found." >&2
+    echo "[otbr-radio]        Check that the BRD2703 is connected and USB-IP is attached." >&2
```
cleithner-comcast: Easy for these messages to become out of date. I would avoid using specifics like models or brands in comments or logs.

cleithner-comcast (changes requested):

docker/otbr-radio-entrypoint.sh
Lakshmi97-velampati: This container is not managing the host system D-Bus. In this setup, compose.otbr-radio.yaml mounts a Docker named volume at dbus, so the socket directory is private to the otbr-radio/barton containers and not a bind mount of the host’s dbus. The otbr-radio container intentionally owns that private dbus-daemon lifecycle, so removing stale pid/socket files in otbr-radio-entrypoint.sh is cleanup of its own bus state, not host state.

If you want to be extra safe/clear, we can move this to a dedicated path like /var/run/otbr-dbus so it’s obvious this is not the host system bus.

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: If RADIO_DEVICE is unset, Compose substitutes the default ttyACM0 in this devices: mapping. Docker resolves that host path when creating the otbr-radio container, so if ttyACM0 does not exist on the host, container creation fails before the entrypoint runs. In other words, this does not degrade gracefully today: the runtime check in otbr-radio-entrypoint.sh only runs after Docker has already attached a real device. That’s why this overlay is intentionally opt-in and should only be used once a valid radio device path exists (locally or via USB/IP).

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: USB/IP only solves serial device access; it does not put OTBR on the host network stack. network_mode: host is still needed so wpan0 exists in the host namespace, OTBR’s NAT64 iptables rules affect real traffic, and mDNS/Avahi multicast reaches the LAN. A custom Docker bridge would isolate all of that inside the container.

tleacmcsa: what about when multiple people are simultaneously using a shared build server (which is how most of us work today)?  There can't be multiple wpan0 interfaces in the host network configuration.  OTBR's NAT64 rules should be able to work on the docker container's NAT network, right?

docker/compose.otbr-radio.yaml
tleacmcsa: again, please consider the common use case whereby we are sharing access to a remote build machine. My USBIP device may connect as /dev/ttyACM8 (for example).

docker/otbr-radio-entrypoint.sh
tleacmcsa: lets be sure the docker containers we use do not connect to the hosts dbus.  They should be starting their own dbus daemon.

docs/THREAD_BORDER_ROUTER_SUPPORT.md
Lakshmi97-velampati: updated the devcontainer.json file to include otbr-radio by default.

docker/otbr-radio-entrypoint.sh
Lakshmi97-velampati: updated

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: Good point. I changed this to use a per-user bridge network instead of network_mode: host.

That avoids wpan0 conflicts on shared build servers, keeps OTBR/NAT64 isolated to the container namespace, and still lets Barton talk to otbr-agent over the private D-Bus volume. so multiple users can run this simultaneously without colliding.

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: Fixed this. I removed the hard-coded devices: mapping and mount dev instead, so Docker no longer fails at container-create time if the radio is absent or appears at a different path like /dev/ttyACM8.

RADIO_DEVICE is now set explicitly by the user, and the entrypoint validates it at runtime with a clear error if it is unset or the device path does not exist. That covers the shared remote build machine case much better.

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: Updated

docker/otbr-radio-entrypoint.sh
Lakshmi97-velampati: Updated this to make it explicit. The containers now use a private D-Bus socket directory at /var/run/otbr-dbus, not /var/run/dbus, and otbr-radio starts its own private dbus-daemon bound to that socket. Barton connects to that private bus via DBUS_SYSTEM_BUS_ADDRESS, so neither container depends on or talks to the host system D-Bus.

docker/compose.otbr-radio.yaml
cleithner-comcast: I just want to point out that the barton service has a custom ipv6 docker network already to properly segment service composition from the host network. Is there problems with just adding:

networks:
            - barton-ip6net

to the new services?

tleacmcsa:
Be sure to take this PR out of draft when you are ready for re-review.

docker/compose.otbr-radio.yaml
```diff
@@ -0,0 +1,148 @@
...
+    # This grants Barton access to the otbr-agent D-Bus API running in the
+    # otbr-radio container without any code changes, while keeping Barton off
+    # the host's system bus.
+    # In devcontainer mode, this file is layered into the compose stack via
+    # devcontainer.json, so this barton service fragment is merged in there.
+    barton:
+        volumes:
+            - dbus-socket:/var/run/otbr-dbus
+        environment:
+            - DBUS_SYSTEM_BUS_ADDRESS=unix:path=/var/run/otbr-dbus/system_bus_socket
```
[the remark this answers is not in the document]
Lakshmi97-velampati: Done

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: update with barton-ip6net

docker/Dockerfile.otbr-radio
```diff
@@ -0,0 +1,188 @@
...
+# otbr-agent — OpenThread Border Router Agent
+#
+# Built from upstream ot-br-posix, with Silicon Labs Gecko SDK vendor transport
+# files layered in to enable CPC/multi-PAN support for `spinel+cpc://`.
+###############################################################################
+ARG GSDK_GIT_BRANCH="gsdk_4.4"
+ARG GSDK_GIT_REV="59d2160fe448b13730a91c0f019a8b4c53c43eb2"
+ARG OTBR_GIT_TAG="731e115257d295a1ec64ea2f60e17a43e43892f6"
+
+RUN cd /tmp && \
```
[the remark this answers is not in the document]
Lakshmi97-velampati: @copilot updated this to OTBR_GIT_REV

cleithner-comcast (dismissed):

rchowdcmcsa (dismissed):

tleacmcsa:
@cleithner-comcast and @kfundecmcsa : you have comments that are still open.  Please review the responses/changes and mark them resolved if satisfied.
@Lakshmi97-velampati same for you, but for copilot's comments.  Please let me know if its ready for me to try out.

kfundecmcsa (dismissed):

Lakshmi97-velampati:
> @cleithner-comcast and @kfundecmcsa : you have comments that are still open. Please review the responses/changes and mark them resolved if satisfied. @Lakshmi97-velampati same for you, but for copilot's comments. Please let me know if its ready for me to try out.

@tleacmcsa yes, you can try this.

cleithner-comcast:
Marked comments resolved

docs/THREAD_BORDER_ROUTER_SUPPORT.md
```diff
@@ -1,3 +1,295 @@
...
+# Start the USB-IP daemon (listens on TCP port 3240)
+sudo usbipd -D
+
+# Confirm that port 3240 is listening
+netstat -an | grep 3240
+
+# Due to firewall restrictions on the laptop, the remote server cannot connect
+# to port 3240 directly. Use an SSH reverse tunnel to work around this.
+# Keep this terminal open while working with the device.
+ssh -R 3240:localhost:3240 <username>@<remote_ip>
```
tleacmcsa: This will only work for one remote developer at a time on the server.  The --tcp-port option should probably be used along with a base port number plus the user's UID.

docker/otbr-radio-entrypoint.sh
```diff
@@ -0,0 +1,178 @@
...
+# before starting a new private dbus-daemon bound to the explicit address.
+###############################################################################
+echo "[otbr-radio] Starting private D-Bus system bus at ${DBUS_SYSTEM_BUS_ADDRESS}..."
+mkdir -p "${DBUS_DIR}"
+if [ -S "${DBUS_SOCKET_PATH}" ]; then
+    echo "[otbr-radio] Removing stale D-Bus socket ${DBUS_SOCKET_PATH}..."
+    rm -f "${DBUS_SOCKET_PATH}"
+fi
+dbus-daemon --system --fork --nopidfile --address="${DBUS_SYSTEM_BUS_ADDRESS}"
+echo "[otbr-radio] Private D-Bus started."
```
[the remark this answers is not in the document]
Lakshmi97-velampati: we are removing the stale D-Bus socket paths before starting the D-Bus

dockerw
```diff
@@ -69,12 +73,24 @@ EXTRA_VOLUMES="${DEV_VOL}"
...
+# background before opening the barton shell.  The container runs cpcd and
+# otbr-agent and shares its D-Bus socket with the barton service.
+# Inject DBUS_SYSTEM_BUS_ADDRESS so Barton automatically connects to the
+# private bus instead of the host system bus.
+if [ ${#OTBR_RADIO_COMPOSE[@]} -gt 0 ]; then
+    notice "Starting otbr-radio container (cpcd + otbr-agent)..."
+    docker compose -f "${COMPOSE_FILE}" "${OTBR_RADIO_COMPOSE[@]}" up -d otbr-radio
+    notice "otbr-radio container started in the background."
+    EXTRA_ENV="${EXTRA_ENV} -e DBUS_SYSTEM_BUS_ADDRESS=unix:path=/var/run/otbr-dbus/system_bus_socket"
+fi
```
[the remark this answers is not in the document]
Lakshmi97-velampati: As part of the current changes, we are removing any existing (stale) D-Bus socket paths before starting D-Bus. Based on this, my understanding is that stale sockets should not persist and cause connection failures.

However, the concern about the container exiting early (e.g., due to an invalid/missing RADIO_DEVICE) still seems valid, as the socket may not be actively served in that case.

@tleacmcsa , could you please confirm if additional validation (such as checking container health or socket availability) is still required here?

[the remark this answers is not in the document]
tleacmcsa: i dont think additional validation is needed yet. I havent had a chance to try these changes out myself, but hope to today.  I'm in favor of getting it basically working, then potentially adding any additional validation if needed.

docker/compose.otbr-radio.yaml
```diff
@@ -0,0 +1,148 @@
...
+    # simulated/default OTBR flows keep working unless the private socket is
+    # explicitly selected by setting DBUS_SYSTEM_BUS_ADDRESS in the compose
+    # environment before startup.
+    # In devcontainer mode, this file is layered into the compose stack via
+    # devcontainer.json, so this barton service fragment is merged in there.
+    barton:
+        volumes:
+            - dbus-socket:/var/run/otbr-dbus
+        environment:
+            - DBUS_SYSTEM_BUS_ADDRESS=${DBUS_SYSTEM_BUS_ADDRESS:-unix:path=/var/run/dbus/system_bus_socket}
```
tleacmcsa: Why isnt this the otbr-dbus?

tleacmcsa (commented):
This generally worked with the exception that I had to change the DBUS_SYSTEM_BUS_ADDRESS value set in compose.otbr-radio.yaml to point to the otbr-dbus path.

Thread subsystem worked, but I could not commission a Matter+Thread device since there is no bluez/Bluetooth support.  We need to add bluez in the same container as otbr-agent since they both can use the same radio for Thread and bluetooth.  Its fine with me if the bluetooth/bluez part comes in a follow-up PR though.

To confirm, were you able to commission a Matter+Thread device in this configuration? How did you decide it was successful, the Thread subsystem showing ready?

docker/compose.otbr-radio.yaml
Lakshmi97-velampati: unix:path=/var/run/dbus/system_bus_socket fallback is the safe default for the no-hardware case, and dockerw -T (or the devcontainer environment when hardware is configured) overrides it to the private bus.

docs/THREAD_BORDER_ROUTER_SUPPORT.md
Lakshmi97-velampati: will update it in next commit

docker/compose.otbr-radio.yaml
tleacmcsa: FYI, i am only testing the devcontainer method (as I expect most users of this would since we almost always work in the IDE).  I did not see how to get this working without hacking on this yaml file to change it to the otbr-dbus.  Please try that out and be sure to test with devcontainer.

Lakshmi97-velampati: Sure will try it.

docker/Dockerfile.otbr-radio
```diff
@@ -0,0 +1,188 @@
...
+###############################################################################
+# cpcd — Co-Processor Communication Daemon
+#
+# cpcd manages the USB serial link to the physical radio using the CPC
+# multiplexing protocol. otbr-agent connects to it via the spinel+cpc:// URI
+# scheme instead of directly to the serial port.
+#
+# Source: https://github.com/SiliconLabs/cpc-daemon
+###############################################################################
+ARG CPCD_GIT_TAG="v4.5.2"
```
tleacmcsa: this should match the gsdk we use for the radio, which is currently "v4.4.6"

Lakshmi97-velampati:
> This generally worked with the exception that I had to change the DBUS_SYSTEM_BUS_ADDRESS value set in compose.otbr-radio.yaml to point to the otbr-dbus path.
> 
> Thread subsystem worked, but I could not commission a Matter+Thread device since there is no bluez/Bluetooth support. We need to add bluez in the same container as otbr-agent since they both can use the same radio for Thread and bluetooth. Its fine with me if the bluetooth/bluez part comes in a follow-up PR though.
> 
> To confirm, were you able to commission a Matter+Thread device in this configuration? How did you decide it was successful, the Thread subsystem showing ready?

Thanks for testing this and for adding the BLE support.

Apologies, I haven’t validated the commissioning flow yet. My checks were limited to confirming that the otbr-agent is running in the new container.

I’ll verify the commissioning flow now with the BLE changes you’ve added and confirm the results.

docs/THREAD_BORDER_ROUTER_SUPPORT.md
Lakshmi97-velampati: you have added this as part of helper scripts

docker/Dockerfile.otbr-radio
Lakshmi97-velampati: Done

.devcontainer/devcontainer.json
```diff
@@ -53,7 +54,11 @@
         // when loaded via GObject introspection in Python. Without this, VS Code's pytest
         // discovery crashes because ASAN kills the process. This is harmless for non-ASAN
         // processes — they simply ignore the option.
-        "ASAN_OPTIONS": "verify_asan_link_order=0"
+        "ASAN_OPTIONS": "verify_asan_link_order=0",
+
+        // Disable debuginfod URL lookups — GDB hangs trying to download symbols from
+        // the Ubuntu debuginfod server when this is set.
+        "DEBUGINFOD_URLS": ""
```
cleithner-comcast: This is already been pushed into the Dockerfile, so you don't need this anymore.

core/src/subsystems/matter/Matter.cpp
```diff
@@ -356,6 +360,53 @@ bool Matter::Init(uint64_t accountId, std::string &&attestationTrustStorePath, c
     return result;
 }
 
+uint32_t Matter::ResolveBleAdapterId()
+{
+    uint32_t adapterId = BLE_CONTROLLER_ADAPTER_ID_DEFAULT;
+    const char *env = getenv(BLE_CONTROLLER_ADAPTER_ID_ENV);
```
cleithner-comcast: FWIW we haven't really used environment variables in our code. We've historically used runtime properties for this. Is this a pattern we want to start adopting? I'm of the mindset of no. I would prefer if the reference app translated environment variable or CLI arguments to properties.

.devcontainer/devcontainer.json
```diff
@@ -7,7 +7,8 @@
     "initializeCommand": "docker/setupDockerEnv.sh",
     "dockerComposeFile": [
         "../docker/compose.yaml",
-        "../docker/compose.devcontainer.yaml"
+        "../docker/compose.devcontainer.yaml",
+        "../docker/compose.otbr-radio.yaml"
```
cleithner-comcast, line 11: The compose uses hierarchical approach but this is the first time two entries in the array both have Dockerfiles. How does that work?

Lakshmi97-velampati: The two Dockerfiles target different services -- compose.yaml builds the barton service (from [Dockerfile]) and compose.otbr-radio.yaml builds the otbr-radio service (from [Dockerfile.otbr-radio]). Docker Compose merges by service name, so there's no ambiguity — each service has exactly one build: block after the merge. This is standard multi-service compose behavior, not a hierarchical override of the same service's build.

cleithner-comcast: Understood, thank you.

core/src/subsystems/matter/Matter.cpp
Lakshmi97-velampati: Done

.devcontainer/devcontainer.json
Lakshmi97-velampati: updated

cleithner-comcast (dismissed):

api/c/public/barton-core-properties.h
```diff
@@ -238,5 +238,14 @@ G_BEGIN_DECLS
 #define B_CORE_BARTON_MATTER_SPAKE2P_VERIFIER                                                                \
     B_CORE_BARTON_PREFIX B_CORE_MATTER_PREFIX "spake2pVerifier"
 
+/**
+ * MATTER_BLE_HCI_INDEX: (value "device.matter.bleAdapterId")
+ *
+ * BLE adapter index (e.g. 0 for hci0, 1 for hci1).
+ * Set by the reference app from the BARTON_BLE_ADAPTER_ID environment variable.
```
cleithner-comcast: The public interface should not refer to the reference app. The reference app is an example client that sits on top of this public interface.

core/src/subsystems/matter/Matter.cpp
```diff
@@ -399,10 +400,24 @@ uint32_t Matter::ResolveBleAdapterId()
...
 
+            // Accept only if at least one digit was parsed and the remainder
+            // is purely whitespace/newline (reject partial parses like "1abc").
             if (endPtr != buf)
             {
-                adapterId = static_cast<uint32_t>(val);
-                icInfo("Using BLE adapter hci%u from %s", adapterId, BLE_CONTROLLER_ADAPTER_ID_FILE);
+                while (*endPtr == ' ' || *endPtr == '\t' || *endPtr == '\n' || *endPtr == '\r')
+                {
+                    endPtr++;
```
cleithner-comcast: Doing pointer iteration is very old school and error prone. This seems like a job for regex to me.

docs/log.txt
cleithner-comcast, line 1: Was this supposed to be checked in?

cleithner-comcast (changes requested):
Pretty close to approval for me.

docs/log.txt
Lakshmi97-velampati: mistakenly committed. Removed this file.

core/src/subsystems/matter/Matter.cpp
Lakshmi97-velampati: Done.

api/c/public/barton-core-properties.h
Lakshmi97-velampati: Updated the comment.

cleithner-comcast (approved):

kfundecmcsa (approved):

tleacmcsa (changes requested):
Reviewed the remote-radio feature. The opt-in CLI path (`dockerw -T`) is wired correctly, but several changes touch shared/core files outside the opt-in gate and can affect everyone, not just remote-radio users. Requesting changes before merge.

### 🔴 Critical — devcontainer makes the "optional" radio feature non-optional, and breaks D-Bus by default

`.devcontainer/devcontainer.json` unconditionally adds `docker/compose.otbr-radio.yaml` to `dockerComposeFile` for every devcontainer user. That overlay's `barton` fragment merges in:

```yaml
environment:
    - DBUS_SYSTEM_BUS_ADDRESS=${DBUS_SYSTEM_BUS_ADDRESS:-unix:path=/var/run/otbr-dbus/system_bus_socket}
```

For anyone without `RADIO_DEVICE`/`RADIO_PORT` set (i.e. almost everyone), the `otbr-radio` container will fail validation and never create that socket, but `barton`'s D-Bus address now points at it anyway — breaking anything that talks over D-Bus (simulated Thread, Zigbee, etc.) by default. It also forces every devcontainer user to build/start a new `privileged: true` container and mounts `/proc/1/ns/net` into `barton` unconditionally. The CLI path (`dockerw -T`) gates this correctly (`DBUS_SYSTEM_BUS_ADDRESS` is only injected when `-T` is passed) — the devcontainer path needs the same gating.

### 🟠 Main image drops the D-Bus policy the existing simulated-OTBR workflow depends on

`docker/Dockerfile` removes `COPY otbr-agent.conf /etc/dbus-1/system.d/`, and `otbr-agent.conf` itself was fully repurposed into a private-bus config now only shipped in `Dockerfile.otbr-radio`. `scripts/start-simulated-otbr.sh` still runs `otbr-agent -I wpan0 ...` against the regular system bus and will now fail to acquire its bus name under the default-deny system policy — silently breaking simulated Thread testing, unrelated to the new feature. Please either restore the needed policy for the main image or remove/update the now-broken script.

### 🟡 Global Matter/BLE behavior changes not scoped to CPC radio use

- `third_party/.../0003-fix-ble-support-LE-only-adapters-...patch` changes `BluezObjectManager::OnObjectAdded` for **all** Matter builds: was `Class != 0`, now `Class != 0 || Powered == true`. This applies unconditionally to all adapter types, not just LE-only/CPC adapters, and could reintroduce the exact "adapter not fully initialized" race the original check guarded against for ordinary dual-mode adapters. Please double-check this doesn't regress commissioning on standard BLE hardware.
- `OpenThreadClient.cpp`: `ATTACH_WAIT_SECONDS` raised 25 → 120 and a new dataset-retry loop, both applied globally. Justified for CPC's shared BLE/Thread radio, but means genuine failures in simulated/non-radio setups now take up to 2 minutes to surface instead of failing fast (slower CI/test feedback).
- `Matter::ResolveBleAdapterId()` adds a property/file lookup to every `Matter::Init()` call; fallback to `hci0` looks safe and low risk — just flagging the added code path runs for everyone.

### 🟢 Lower risk / worth noting

- `bluez` added to the main (non-radio) image's apt install list — image bloat for everyone, not just radio users.
- `barton-core-reference-io.c` makes the log pipe non-blocking globally; reasonable deadlock fix, but means log lines can now be silently dropped under logging pressure for all reference-app users.
- `docker/version` bump to 3.0 forces a rebuild for everyone — expected cost given the Dockerfile changes, just flagging it.

---

### Architectural suggestion

The root cause of most of the above is that "simulated OTBR" and "real-radio OTBR" are two disjoint code paths today: simulated `otbr-agent` runs inside the `barton` container on the system bus; real-radio `otbr-agent` runs in a separate container on a private bus, only wired up when explicitly opted into. That's what's forcing the devcontainer to special-case D-Bus wiring and strip the system-bus policy, causing the regressions above.

Suggestion: always run a single `otbr` container whenever Thread is enabled, with the entrypoint choosing the RCP transport based on env:
- No `RADIO_DEVICE`/`RADIO_PORT` → start `ot-rcp` locally, `spinel+hdlc+forkpty://` (today's simulated behavior).
- `RADIO_DEVICE`/`RADIO_PORT` set → start `cpcd`, `spinel+cpc://` against real hardware.

Either way it exposes the same private-bus D-Bus socket to `barton` the same way, always — so `DBUS_SYSTEM_BUS_ADDRESS` is never pointed at a socket that might not exist, `otbr-agent` never needs to live inside the main `barton` image, and simulated vs. real-hardware testing exercise the identical container/D-Bus plumbing (just a different backend), removing the drift that caused this PR's regressions. BLE-over-CPC (bridge/btattach/bluetoothd/host-netns) should stay conditional on `RADIO_DEVICE`/`RADIO_PORT`, since it's meaningless in simulated mode. Tradeoffs to plan for: the `otbr` image's build cost (cpcd + gecko_sdk + ot-br-posix from source) would now land on every build rather than just radio users, so it should be cached/pushed rather than rebuilt from scratch, and `barton` should gain a proper `depends_on: otbr: condition: service_healthy` instead of relying on incidental startup ordering.

---

Requesting changes to address the devcontainer D-Bus regression and the simulated-OTBR D-Bus policy removal before merge; happy to discuss the architectural suggestion separately if it's out of scope for this PR.

PR #202, opened 2026-04-08
feat(sbmd): add IKEA TIMMERFLOTTE sensor support with vendor-specific claiming

kfundecmcsa:
Add SBMD-based drivers for temperature, humidity, and combined
temperature/humidity sensors (IKEA TIMMERFLOTTE). This introduces
vendor-specific claiming so a driver with vendorId/productId gets
priority over generic device-type drivers, preventing a single-type
driver from incorrectly claiming a multi-function vendor device.

## Commits
See individual commit messages for full details.
- **docs(openspec): add temp/humidity sensor change** — design, specs,
  and tasks for the change
- **build(sbmd): support versioned schema directories** — support
  multiple schema versions so older specs remain valid as the schema
  evolves; patch mqjs to actually report compile errors, fix latent
  const usage in contact-sensor.sbmd
- **feat(sbmd): add vendor/product ID claiming and endpoint fallback**
  — vendor-specific drivers get priority over generic device-type
  drivers, cross-endpoint resource binding for multi-function devices
- **feat(sbmd): add temperature, humidity, and IKEA TIMMERFLOTTE sensor
  drivers** — SBMD specs, integration tests, and mock devices

OpenSpec-Change: sbmd-temperature-humidity-spec
Refs: BARTON-350


core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,39 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+        if (!deviceDataCache ||
+            !deviceDataCache->EndpointHasServerCluster(outEndpointId, chip::app::Clusters::Descriptor::Id) ||
+            deviceDataCache->EndpointHasServerCluster(outEndpointId, clusterId))
+        {
+            return true;
+        }
+
+        // The mapped endpoint doesn't host this cluster (e.g. composite devices like
+        // temperature/humidity sensors where clusters span multiple Matter endpoints but
+        // are grouped under one SBMD endpoint). Fall back to cluster-based lookup.
```
cleithner-comcast: How does a cluster span multiple endpoints? I'm not understanding this comment. It also looks to me like the example used (temp/humidity sensor) has explicitly mapped clusters/attributes according to the spec so I'm not sure when ResolveEndpointForCluster would go through this branch.

Could it be this is saying "when multiple Matter endpoints host the same cluster and they are not mapped to an SBMD endpoint"? In that case, is the old cluster lookup (first endpoint with the cluster I think?) correct or is erroring correct?

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -138,43 +139,60 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
             {
-                icDebug("Device claimed: endpoint %d has matching device type 0x%04x", endpointId, deviceTypeId);
-                return true;
+                icDebug("Device not claimed: missing required device type 0x%04x", deviceType);
+                return false;
             }
         }
+
+        icDebug("Device claimed: all required device types present");
+        return true;
```
cleithner-comcast: the branching model in this routine is a bit sloppy due to many early returns combined  with conditionals and looping. Consider a single return statement at the end.

testing/test/temperature_humidity_sensor_test.py
cleithner-comcast, line 1: I would probably split this into three test files. Having multiple device integration tests in one file seems unnecessary to me.

cleithner-comcast (commented):
Generally good with it, some open questions.

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
```diff
@@ -0,0 +1,51 @@
...
+# Barton device class mapping
+bartonMeta:
+  deviceClass: "humiditySensor"
+  deviceClassVersion: 1
+
+# Matter device type support
+matterMeta:
+  deviceTypes:
+    - 0x0307  # Humidity Sensor
+  revision: 1
```
tleacmcsa: This revision comes from the Matter Device Library specification and indicates the version of the device type this driver supports. This should be 3 (please find it in the spec to confirm).

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
```diff
@@ -0,0 +1,51 @@
...
+          read:
+            attribute:
+              clusterId: "0x0405"      # Relative Humidity Measurement cluster
+              attributeId: "0x0000"    # MeasuredValue attribute
+              name: "MeasuredValue"
+              type: "uint16"
+            script: |
+              var value = SbmdUtils.Tlv.decode(sbmdReadArgs.tlvBase64);
+              if (value === null || value === 0xFFFF) {
+                return {output: ''};
```
tleacmcsa, line 55: Should we have better error reporting for these situations (not just this driver)? could track in another ticket i suppose.

core/deviceDrivers/matter/sbmd/specs/temperature-sensor.sbmd
```diff
@@ -0,0 +1,51 @@
...
+# Barton device class mapping
+bartonMeta:
+  deviceClass: "temperatureSensor"
+  deviceClassVersion: 1
+
+# Matter device type support
+matterMeta:
+  deviceTypes:
+    - 0x0302  # Temperature Sensor
+  revision: 1
```
tleacmcsa: please check this revision too

core/deviceDrivers/matter/sbmd/specs/temperature-sensor.sbmd
```diff
@@ -0,0 +1,51 @@
...
+  minSecs: 1
+  maxSecs: 3600
+
+endpoints:
+  - id: "1"
+    profile: "sensor"
+    profileVersion: 2
+    resources:
+      - id: "temperature"
+        type: "com.icontrol.temperature"
```
tleacmcsa, line 41: This is an existing/used resource type.  see zigbeeDriverCommon.c.  Make sure the units, etc. match.  cant have one temperature resource working differently than another.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
```diff
@@ -0,0 +1,74 @@
...
+  deviceClass: "temperatureHumiditySensor"
+  deviceClassVersion: 1
+
+# Matter device type support
+matterMeta:
+  deviceTypes:
+    - 0x0302  # Temperature Sensor
+    - 0x0307  # Humidity Sensor
+  deviceTypeMatch: "all"
+  revision: 1
```
tleacmcsa: it does not make sense to have a shared revision across device types.

tleacmcsa (commented):
partial review... just the sbmd files so far.

testing/mocks/devices/matterjs/src/HumiditySensorDevice.js
```diff
@@ -0,0 +1,110 @@
...
+ *   - Initial state: 50.00% (raw value 5000)
+ *
+ * Can be run directly:  node HumiditySensorDevice.js --passcode ... --discriminator ...
+ */
+
+import {Endpoint} from "@matter/main";
+import {
+    HumiditySensorDevice as MatterHumiditySensorDevice,
+    HumiditySensorRequirements,
+} from "@matter/main/devices";
```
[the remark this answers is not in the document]
kfundecmcsa: I think it's fine

testing/mocks/devices/matterjs/src/DoorLockDevice.js
```diff
@@ -56,50 +56,47 @@ export class DoorLockDevice extends VirtualDevice {
...
+        await this.endpoints[0].act(async (agent) => { agent.doorLock.state.lockState = DoorLock.LockState.Locked; });
 
         return { lockState: "locked" };
     }
 
     async handleUnlock() {
-        await this.endpoint.act(async (agent) => {
-            agent.doorLock.state.lockState = DoorLock.LockState.Unlocked;
-        });
+        await this.endpoints[0].act(async (agent) => { agent.doorLock.state.lockState = DoorLock.LockState.Unlocked; });
```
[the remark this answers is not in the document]
kfundecmcsa: This is done by the clang formatter in the pre commit hook, leaving it for now

testing/test/temperature_humidity_sensor_test.py
kfundecmcsa: good call, done

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
kfundecmcsa: Should have caught this assumption AI made. Confirmed in 1.5.1 spec this is version 3 (as well as temperature)

core/deviceDrivers/matter/sbmd/specs/temperature-sensor.sbmd
kfundecmcsa: Confirmed this should be 3 as of spec 1.5.1

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: makes sense, removed

tleacmcsa: im not sure simply removing it is the right choice either.

core/deviceDrivers/matter/MatterDevice.cpp
kfundecmcsa: The comments in this function could have been worded better. I redid the comment structure and wording to make this more clear. Please take a look and see if it makes more sense now

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
kfundecmcsa: Ticket logged as BARTON-366

core/deviceDrivers/matter/MatterDeviceDriver.cpp
kfundecmcsa: agreed, done

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: Right, good point.

What if we restructured `deviceTypes` so each entry is a mapping that pairs the device type ID with its revision?

```yaml
deviceTypes:
  - id: 0x0302
    revision: 3
  - id: 0x0307
    revision: 3
```

This would be a schema-wide change affecting all existing SBMD specs, so might be best tackled in a separate ticket. Worth noting that `revision` is currently unused at runtime; it's parsed and stored but never consumed. For now, I propose leaving it as `revision: 3` with a TODO comment explaining the situation.

Thoughts?

core/deviceDrivers/matter/sbmd/specs/temperature-sensor.sbmd
kfundecmcsa: Good catch, the mappers were dividing by 100 and outputting decimal strings like `"23.50"` instead of the raw integer `"2350"` that the Zigbee side uses. Fixed. Did the same for humidity to keep it consistent.

cleithner-comcast (dismissed):

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
tleacmcsa: Thought about this more. I now DO think that we should remove it until we think we need it. Sorry for the thrash.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
```diff
@@ -0,0 +1,78 @@
+# Temperature Humidity Sensor SBMD Specification
+# Maps composite Matter device with both temperature and humidity sensor
```
tleacmcsa: "composite" has meaning in the Matter spec, and it does not match this usage.  Let's call this "combined".  A composite device uses PartsList attribute on the Descriptor cluster to "string together" endpoints and their device types into a complex device composed of explicit parts from the standard.  Like a refrigerator should have temperature sensors, lights, etc.   This Ikea device is just one device with two things in it.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
```diff
@@ -0,0 +1,78 @@
...
+schemaVersion: "1.0"
+# Driver version for this specification
+driverVersion: "1.0"
+# Human-readable driver name
+name: "Temperature Humidity Sensor"
+scriptType: "JavaScript"
+
+# Barton device class mapping
+bartonMeta:
+  deviceClass: "temperatureHumiditySensor"
```
tleacmcsa: I'm unsure about this device class, so we should have a discussion.  There is not necessarily a 1:1 mapping between the end device type and our device class.  Are we making a Barton device class this specific, or are we making something like an air environmental sensor or some such?  could be defined to have optional resources that might include other stuff if the end device has it.

core/deviceDrivers/matter/sbmd/specs/ikea-timmerflotte.sbmd
```diff
@@ -0,0 +1,78 @@
...
+
+# Barton device class mapping
+bartonMeta:
+  deviceClass: "temperatureHumiditySensor"
+  deviceClassVersion: 1
+
+# Matter device type support
+matterMeta:
+  deviceTypes:
+    - 0x0302  # Temperature Sensor
```
tleacmcsa, line 23: It just occurred to me that another approach would be to have this be the ikea TIMMERFLOTTE driver that binds to the vendor and product ids rather than these two combined device types.  What are the odds that someone else will make a sensor that combines these two things?  Will we ever need drivers for specific vendor devices?   probably...  Might be able to ditch the deviceTypeMatch stuff if we went down that route.

core/deviceDrivers/matter/sbmd/SbmdSpec.h
```diff
@@ -207,6 +207,7 @@ namespace barton
         std::vector<uint16_t> deviceTypes;
         uint32_t revision;
         std::vector<uint32_t> featureClusters; // Optional: cluster IDs to get feature maps from
+        std::string deviceTypeMatch = "any";   // "any" (default) or "all" for composite device matching
```
tleacmcsa: composite->combined ?  just a reminder that we should think about the terminology everwhere we added it

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,45 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+{
+    if (sbmdEndpointIndex.has_value())
+    {
+        if (!GetEndpointForSbmdIndex(sbmdEndpointIndex.value(), outEndpointId))
+        {
+            return false;
+        }
+
+        // One SBMD endpoint maps to one Matter endpoint, but the SBMD endpoint's
+        // resources may reference clusters that live on different Matter endpoints
```
tleacmcsa: Isnt this contradictory?  First it says a 1:1 mapping of SBMD endpoints to Matter endpoints, then says an SBMD endpoint's resources could refer to other Matter endpoints.

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,45 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+        // (e.g. a temperature-humidity sensor has one SBMD endpoint with both
+        // temperature and humidity resources, but the clusters are on separate
+        // Matter endpoints). To resolve the correct Matter endpoint: first check if the
+        // mapped endpoint hosts the requested cluster, and if not, fall back to
+        // scanning all endpoints.
+
+        // Use the mapped endpoint if it hosts this cluster (or if we can't
+        // verify because cache data isn't available yet).
```
tleacmcsa: Is there ever a case whereby we dont have a cache available yet?

core/deviceDrivers/matter/MatterDeviceDriver.h
```diff
@@ -100,6 +100,20 @@ namespace barton
 
         virtual bool ClaimDevice(const DeviceDataCache *deviceDataCache);
 
+        /**
+         * @brief Returns true if this driver requires ALL advertised device types to match
+         * before it will claim a device, rather than matching on any single type.
+         *
+         * Composite drivers model multi-endpoint devices that expose several distinct
```
tleacmcsa: Consider the Matter specification's definition of a composite device in this description to make sure we dont design an API that wont make sense when we add real support for true composite devices.

cleithner-comcast:
Should this be draft?

kfundecmcsa:
> Should this be draft?

yes, done

core/deviceDrivers/matter/sbmd/specs/ikea-timmerflotte.sbmd
```diff
@@ -0,0 +1,88 @@
...
+        mapper:
+          read:
+            alias: "measuredHumidity"
+            script: |
+              var value = SbmdUtils.Tlv.decode(sbmdReadArgs.tlvBase64);
+              // 0xFFFF: Matter null for uint16 MeasuredValue
+              if (value === null || value === 0xFFFF) {
+                return {output: ''};
+              }
+              return {output: value.toString()};
```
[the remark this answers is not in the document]
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
```diff
@@ -0,0 +1,57 @@
...
+        mapper:
+          read:
+            alias: "measuredHumidity"
+            script: |
+              var value = SbmdUtils.Tlv.decode(sbmdReadArgs.tlvBase64);
+              // 0xFFFF: Matter null for uint16 MeasuredValue
+              if (value === null || value === 0xFFFF) {
+                return {output: ''};
+              }
+              return {output: value.toString()};
```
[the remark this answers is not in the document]
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/schema/v2/sbmd-spec-schema-v2.1.json
```diff
@@ -0,0 +1,476 @@
...
+        },
+        "vendorId": {
+          "type": ["integer", "string"],
+          "description": "Matter vendor ID for vendor-specific claiming (hex or decimal)"
+        },
+        "productId": {
+          "type": ["integer", "string"],
+          "description": "Matter product ID for vendor-specific claiming (hex or decimal)"
+        }
+      },
```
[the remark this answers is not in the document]
kfundecmcsa: Added dependentRequired to enforce the "both or neither" constraint at schema validation time. Dropped the minimum/maximum bounds since they only apply to the integer branch of the type union (hex strings like "0xFFFFFF" would bypass them), and no other ID fields in the schema use range constraints. The parser's range check (added in this same commit) is the authoritative backstop for uint16 overflow.

core/deviceDrivers/matter/MatterDeviceDriver.h
```diff
@@ -100,6 +100,36 @@ namespace barton
...
+         */
+        virtual uint16_t GetSupportedVendorId() const { return 0; }
+
+        /**
+         * @brief Get the product ID this driver targets.
+         * @return 0: no product constraint. As with vendorId, Matter reserves 0x0000.
+         * @return non-zero: ClaimDevice will match this value against the device's
+         *         BasicInformation productId alongside the vendorId.
+         */
+        virtual uint16_t GetSupportedProductId() const { return 0; }
```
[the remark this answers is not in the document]
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdParser.cpp
```diff
@@ -418,6 +418,24 @@ bool SbmdParser::ParseMatterMeta(const YAML::Node &node, SbmdMatterMeta &meta)
...
+        return false;
+    }
+
+    if (hasVendorId)
+    {
+        std::string vendorStr = node["vendorId"].as<std::string>();
+        std::string productStr = node["productId"].as<std::string>();
+        meta.vendorId = static_cast<uint16_t>(ParseHexOrDecimal(vendorStr));
+        meta.productId = static_cast<uint16_t>(ParseHexOrDecimal(productStr));
+    }
```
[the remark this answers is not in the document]
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdSpec.h
```diff
@@ -236,6 +236,8 @@ namespace barton
         uint32_t revision;
         std::vector<uint32_t> featureClusters; // Optional: cluster IDs to get feature maps from
         std::vector<SbmdAlias> aliases;        // Named Matter element definitions referenced by resources
+        std::optional<uint16_t> vendorId;
+        std::optional<uint16_t> productId;
```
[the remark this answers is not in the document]
cleithner-comcast: Make it optional and handle any logic safely.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -138,47 +138,92 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
+        uint16_t deviceProductId = 0;
+        uint16_t expectedVendorId = GetSupportedVendorId();
+        uint16_t expectedProductId = GetSupportedProductId();
+
+        if (deviceDataCache->GetVendorId(deviceVendorId) == CHIP_NO_ERROR &&
+            deviceDataCache->GetProductId(deviceProductId) == CHIP_NO_ERROR)
         {
-            continue;
-        }
+            claimed = deviceVendorId == expectedVendorId && deviceProductId == expectedProductId;
```
cleithner-comcast: minor: consider some explicit parentheses just to break up the statement a little more.

core/test/src/sbmdParserTest.cpp
cleithner-comcast, line 1: I can imagine this test shouldn't grow infinite with every sbmd file we make. We should not change anything in this review so as to not introduce scope creep. However, I think we should kick out a story to reorganize this or introduce something a bit more generic.

core/CMakeLists.txt
```diff
@@ -190,10 +193,10 @@ if (BCORE_MATTER)
 
         # Create a custom target that validates SBMD specs
         add_custom_target(validate_sbmd_specs ALL
-            COMMAND ${Python3_EXECUTABLE} ${SBMD_VALIDATOR} ${SBMD_SCHEMA_FILE} ${SBMD_SPEC_FILES}
+            COMMAND ${Python3_EXECUTABLE} ${SBMD_VALIDATOR} ${SBMD_SCHEMA_DIR} ${SBMD_SPEC_FILES}
```
cleithner-comcast, line 197: This will break again when a v3 directory is added (i.e. when there are multiple schema dirs)

testing/mocks/devices/matterjs/src/IkeaTimmerflotteDevice.js
cleithner-comcast, line 1: Should this be renamed to IkeaTimmerflotteDevice.js ?

cleithner-comcast (changes requested):
Copilot has some good suggestions this time around.

openspec/changes/sbmd-temperature-humidity-spec/specs/temperature-humidity-sbmd-drivers/spec.md
```diff
@@ -0,0 +1,75 @@
...
+### Requirement: IKEA TIMMERFLOTTE sensor resources
+The `ikea-timmerflotte.sbmd` spec SHALL define a single SBMD endpoint with profile `sensor` containing two resources: `temperature` and `humidity`.
+
+#### Scenario: Temperature resource read
+- **WHEN** the `temperature` resource is read from a composite temperature-humidity device
+- **THEN** the driver SHALL read Matter cluster `0x0402` (Temperature Measurement), attribute `0x0000` (MeasuredValue, type `int16`), and the mapper script SHALL convert the raw value (in 0.01°C units) to a decimal Celsius string (e.g., raw `2550` → `"25.50"`)
+
+#### Scenario: Humidity resource read
+- **WHEN** the `humidity` resource is read from a composite temperature-humidity device
+- **THEN** the driver SHALL read Matter cluster `0x0405` (Relative Humidity Measurement), attribute `0x0000` (MeasuredValue, type `uint16`), and the mapper script SHALL convert the raw value (in 0.01% units) to a decimal percentage string (e.g., raw `5000` → `"50.00"`)
```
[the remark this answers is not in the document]
kfundecmcsa: yeah the implementation was fixed to fall in line with expectation, but this spec part wasn't updated - done

core/deviceDrivers/matter/sbmd/SbmdSpec.h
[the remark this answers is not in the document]
kfundecmcsa: good catch - done

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,44 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+        // The Descriptor cluster (0x001D) lists which server clusters an
+        // endpoint hosts. We use it to verify the mapped endpoint actually
+        // has the cluster we need. If both checks are false — we have
+        // Descriptor data AND it says the cluster isn't here — fall through
+        // to scan all endpoints for one that does host it.
+        if (!deviceDataCache->EndpointHasServerCluster(outEndpointId, chip::app::Clusters::Descriptor::Id) ||
+            deviceDataCache->EndpointHasServerCluster(outEndpointId, clusterId))
+        {
+            return true;
+        }
```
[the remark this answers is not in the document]
kfundecmcsa: fair - done

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -138,47 +138,92 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
+                continue;
+            }
+
+            std::vector<uint16_t> deviceTypes;
+            if (deviceDataCache->GetDeviceTypes(endpointId, deviceTypes) != CHIP_NO_ERROR)
+            {
+                continue;
+            }
+
+            auto supportedDeviceTypes = GetSupportedDeviceTypes();
```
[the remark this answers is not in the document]
kfundecmcsa: I caught this in the first iteration but forgot it in the second 😞 

done

core/deviceDrivers/matter/MatterDeviceDriver.cpp
kfundecmcsa: yeah good call, done

core/test/src/sbmdParserTest.cpp
kfundecmcsa: agreed - ticket has been created: `BARTON-379`

testing/mocks/devices/matterjs/src/IkeaTimmerflotteDevice.js
kfundecmcsa: I considered that, but went with the generic `TemperatureHumiditySensorDevice` naming because the mock python device is the thing that actually sets the product ID / vendor ID. So the reasoning was - this class sets up the clusters and really doesn't have knowledge of whether it's ikea or whoever. 

But now I'm thinking about it, it's probably better to just bake in the pid/vid into the javascript virtual device and make it a specific ikea device, and the python device won't need to set the vid/pid.

I made this change, please take a look when you can.

core/CMakeLists.txt
kfundecmcsa: Yarp

I updated this to handle all subdirs of the top level schema dir in the new `build` commit. Please take a look when you can.

cleithner-comcast (dismissed):

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,50 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+        // index, but the SBMD endpoint's resources may reference clusters
+        // hosted on other Matter endpoints. Check whether the mapped endpoint
+        // hosts the requested cluster; if not, scan all endpoints.
+
+        // The Descriptor cluster (0x001D) lists which server clusters an
+        // endpoint hosts. We use it to verify the mapped endpoint actually
+        // has the cluster we need. If both checks are false — we have
+        // Descriptor data AND it says the cluster isn't here — fall through
+        // to scan all endpoints for one that does host it.
+        if (!deviceDataCache->EndpointHasServerCluster(outEndpointId, chip::app::Clusters::Descriptor::Id) ||
```
tleacmcsa: I'm curious why we check if there is a descriptor cluster, since it is required per spec.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: After discussion with the team, we decided to drop the composite/combined approach in favor of vendor specific claiming. As such, this comment is moot now.

core/deviceDrivers/matter/sbmd/SbmdSpec.h
kfundecmcsa: After discussion with the team, we decided to drop the composite/combined approach in favor of vendor specific claiming. As such, this comment is moot now.

core/deviceDrivers/matter/MatterDevice.cpp
kfundecmcsa: Good point, I updated this comment to be more clear here.

core/deviceDrivers/matter/MatterDevice.cpp
kfundecmcsa: No there isn't, I removed that check and updated this comment accordingly

core/deviceDrivers/matter/MatterDeviceDriver.h
kfundecmcsa: After discussion with the team, we decided to drop the composite/combined approach in favor of vendor specific claiming. As such, this comment is moot now.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: This device class was dropped in favor of a specific `ikeaTimmerflotte` driver that's claimed via its VID and PID. Does that address your concerns here?

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
tleacmcsa: The point was about terminology, and "composite" is still used in the current PR.  It may be fine and I havent gotten too far yet, but the point about terminology still applies and should be understood.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
tleacmcsa: Not really.  Device Class is a contract in the Barton data model. It dictates what device level resources are required/optional (if any), and what endpoints are required or optional and their endpoint profiles.  We dont want to expose this very detailed device class as a First Class citizen in our data model.  Maybe we start with version 1.0 of an "environmentalSensor" that would have the two optional endpoints, but be different from the existing "airQualitySensor" device class we recently added.

core/deviceDrivers/matter/sbmd/schema/v2/sbmd-spec-schema-v2.1.json
tleacmcsa, line 1: I wonder if we should start maintaining a changelog for these schema updates so we know what changed.  AI work of course.

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
```diff
@@ -0,0 +1,59 @@
...
+schemaVersion: "2.0"
+# Driver version for this specification
+driverVersion: "1.0"
+# Human-readable driver name
+name: "Humidity Sensor"
+scriptType: "JavaScript"
+
+# Barton device class mapping
+bartonMeta:
+  deviceClass: "humiditySensor"
```
tleacmcsa: if we decide on something like "environmentalSensor" device class for the ikea driver, these drivers would use it too.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: Hmm, I thought I dropped all references to "composite" - I'll look it over

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
```diff
@@ -43,6 +43,10 @@ namespace barton
         SpecBasedMatterDeviceDriver(std::shared_ptr<SbmdSpec> spec);
         std::vector<uint16_t> GetSupportedDeviceTypes() override;
 
+        uint16_t GetSupportedVendorId() const override;
+        uint16_t GetSupportedProductId() const override;
```
tleacmcsa, line 47: these ids need to be lists since it is possible that a driver applies to multiple products/vendors.  I'm ok with that being follow on (need a ticket).

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -138,47 +138,92 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
+                        deviceVendorId,
+                        deviceProductId,
+                        expectedVendorId,
+                        expectedProductId);
+            }
+        }
+        else
         {
-            continue;
+            icDebug("Device not claimed: could not read vendor/product ID from cache");
```
tleacmcsa: I think this would be an error, not debug.

core/deviceDrivers/matter/sbmd/schema/v2/sbmd-spec-schema-v2.1.json
kfundecmcsa: good idea - added a CHANGELOG to `core/deviceDrivers/matter/sbmd/schema/CHANGELOG.md`

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
kfundecmcsa: Good call, done - BARTON-387 has been created to capture this

core/deviceDrivers/matter/MatterDeviceDriver.cpp
kfundecmcsa: correct - fixed

core/deviceDrivers/matter/MatterDevice.cpp
kfundecmcsa: Good call, no reason to guard on Descriptor - it's mandatory per spec and always in the cache by this point (ResolveEndpointMap already reads from it successfully before we get here). Simplified to just check if the mapped endpoint hosts the requested cluster directly.

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: yup - found some remaining `composite` references -  all have been removed.

core/deviceDrivers/matter/sbmd/specs/humidity-sensor.sbmd
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/specs/temperature-humidity-sensor.sbmd
kfundecmcsa: I see - this makes sense.

I made this change, please take a look and see what you think.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -138,47 +138,92 @@ MatterDeviceDriver::~MatterDeviceDriver()
...
+                        deviceVendorId,
+                        deviceProductId,
+                        expectedVendorId,
+                        expectedProductId);
+            }
+        }
+        else
         {
-            continue;
+            icError("Device not claimed: could not read vendor/product ID from cache");
```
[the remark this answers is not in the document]
kfundecmcsa: incorrect - an error is approriate here

openspec/changes/archive/2026-05-04-sbmd-temperature-humidity-spec/specs/temperature-humidity-sbmd-drivers/spec.md
```diff
@@ -0,0 +1,75 @@
...
+- **WHEN** the `humidity` resource is read from a composite temperature-humidity device
+- **THEN** the driver SHALL read Matter cluster `0x0405` (Relative Humidity Measurement), attribute `0x0000` (MeasuredValue, type `uint16`), and the mapper script SHALL convert the raw hundredths-of-a-percent value to whole percent as a string (e.g., raw `5000` → `"50"`)
+
+#### Scenario: Null measurement value
+- **WHEN** the MeasuredValue attribute for temperature or humidity is null (0x8000 for int16 or 0xFFFF for uint16)
+- **THEN** the mapper script SHALL return an empty string `""`
+
+### Requirement: Temperature sensor SBMD spec
+The system SHALL include an SBMD spec file `temperature-sensor.sbmd` that defines a `temperatureSensor` device class matching Matter device type `0x0302` (Temperature Sensor) using the default `"any"` match semantics.
+
```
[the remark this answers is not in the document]
kfundecmcsa: the addition of `environmentalSensor` came from comments and was not considered during spec development. If I have to make another commit, I'll update the spec, else I'm ok with the inconsistency to reduce churn.

openspec/changes/archive/2026-05-04-sbmd-temperature-humidity-spec/specs/temperature-humidity-sbmd-drivers/spec.md
```diff
@@ -0,0 +1,75 @@
...
+- **WHEN** a Matter device has an endpoint with device type `0x0302`
+- **THEN** the `temperature-sensor` driver SHALL claim the device
+
+#### Scenario: Temperature resource read
+- **WHEN** the `temperature` resource is read
+- **THEN** the driver SHALL read Matter cluster `0x0402`, attribute `0x0000` (MeasuredValue, `int16`), and the mapper script SHALL return the raw hundredths-of-a-degree value as a string
+
+### Requirement: Humidity sensor SBMD spec
+The system SHALL include an SBMD spec file `humidity-sensor.sbmd` that defines a `humiditySensor` device class matching Matter device type `0x0307` (Humidity Sensor) using the default `"any"` match semantics.
+
```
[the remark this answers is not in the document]
kfundecmcsa: Same as with the temperature-humidity sensor comment - I'll update if another commit is incoming.

cleithner-comcast (dismissed):

core/deviceDrivers/matter/MatterDeviceDriver.h
```diff
@@ -100,6 +100,37 @@ namespace barton
...
+         *        targets an exact device model identified by its Matter BasicInformation
+         *        vendor and product IDs, rather than matching by endpoint device types.
+         *        This is used for devices that need custom handling beyond what a
+         *        generic device-type driver provides. Subclasses override this to
+         *        indicate they target a specific vendor/product pair.
+         *
+         * @return true if the driver targets a specific vendor/product pair
+         * @return false if the driver matches by device type (generic, the default)
+         */
+        virtual bool IsVendorSpecificDriver() const { return false; }
```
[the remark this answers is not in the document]
kfundecmcsa: I think the current implementation is fine - this vip/pid is an sbmd concept, no reason for it to bleed into the base class.

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -255,6 +255,43 @@ bool MatterDevice::GetEndpointForCluster(chip::ClusterId clusterId, chip::Endpoi
...
+        if (!GetEndpointForSbmdIndex(sbmdEndpointIndex.value(), outEndpointId))
+        {
+            return false;
+        }
+
+        // The endpoint map provides a default Matter endpoint for this SBMD
+        // index, but the SBMD endpoint's resources may reference clusters
+        // hosted on other Matter endpoints. Check whether the mapped endpoint
+        // hosts the requested cluster; if not, scan all endpoints.
+        if (deviceDataCache->EndpointHasServerCluster(outEndpointId, clusterId))
```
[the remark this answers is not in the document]
kfundecmcsa: The cache is always populated by this point - ResolveEndpointForCluster is only called during resource binding, which happens after ResolveEndpointMap has already successfully read Descriptor/DeviceTypeList data from the subscription cache. There's no scenario where the Descriptor data is unavailable when we reach this code.

tleacmcsa (dismissed):

cleithner-comcast (approved):

tleacmcsa (approved):

PR #206, opened 2026-04-14
test: suppress test output unless test fails

kfundecmcsa:
- Run all tests in subprocesses, not just requires_matterjs tests
  - Restores output isolation lost when pytest-forked was replaced with
    the custom subprocess runner - non-matterjs tests (e.g.,
    device_descriptor) were running in-process, causing C library log
    output to bleed into the terminal
- Switch pytest capture from `--capture=sys` to `--capture=fd`
  - sys only intercepts Python's sys.stdout; fd captures at the file
    descriptor level, catching C library fprintf output too
- Remove 160-line truncation on failure output so developers can see
  full diagnostics
- Add pytest `--durations=10` to print the 10 slowest tests at the end
  of each run so we can identify and diagnose slow tests

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -267,14 +267,6 @@ def _run_in_subprocess(item):
 
             if not lines:
                 excerpt = "<no output captured from child pytest process>"
-            elif len(lines) > 160:
-                head = "\n".join(lines[:80])
-                tail = "\n".join(lines[-80:])
-                excerpt = (
-                    f"{head}\n\n"
-                    f"... ({len(lines) - 160} lines omitted) ...\n\n"
-                    f"{tail}"
-                )
             else:
                 excerpt = "\n".join(lines)
 
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -119,17 +119,18 @@ def pytest_collection_modifyitems(config, items):
 
 
 # ---------------------------------------------------------------------------
-#  Subprocess isolation for tests that need a fresh process.
+#  Subprocess isolation for all tests.
 #
-#  The Matter SDK is a C++ singleton that cannot reinitialize within a single
-#  process.
+#  Every test is re-invoked in its own subprocess.  This provides:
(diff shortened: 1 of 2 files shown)


cleithner-comcast (changes requested):
Where the change we talked about to make all the tests not output unless failed (and not some tests)?

kfundecmcsa:
> Where the change we talked about to make all the tests not output unless failed (and not some tests)?

Cause I forgot...

I looked into it, it's a bit more complicated than this fix, I'll keep at it

tleacmcsa:
> > Where the change we talked about to make all the tests not output unless failed (and not some tests)?
> 
> Cause I forgot...
> 
> I looked into it, it's a bit more complicated than this fix, I'll keep at it

Should this PR go back to draft?

kfundecmcsa:
> > > Where the change we talked about to make all the tests not output unless failed (and not some tests)?
> > 
> > 
> > Cause I forgot...
> > I looked into it, it's a bit more complicated than this fix, I'll keep at it
> 
> Should this PR go back to draft?

yes - done

cleithner-comcast (dismissed):

testing/conftest.py
```diff
@@ -119,32 +119,31 @@ def pytest_collection_modifyitems(config, items):
 
 
 # ---------------------------------------------------------------------------
-#  Subprocess isolation for tests that need a fresh process.
+#  Subprocess isolation for all tests.
 #
-#  The Matter SDK is a C++ singleton that cannot reinitialize within a single
-#  process.
+#  Every test is re-invoked in its own subprocess.  This provides:
```
tleacmcsa: "re-invoked"?

tleacmcsa (dismissed):

testing/conftest.py
kfundecmcsa: bah, good catch - fixed

cleithner-comcast (approved):

tleacmcsa (approved):

PR #211, opened 2026-04-23
feat(sbmd): seed event-driven resource values from attributes

rchowdcmcsa:
Event-driven resources (those using mapper.event) previously had no initial value until the first event fired. This change adds a seedFrom mapper type that reads the corresponding attribute from the device's attribute cache at commission time and again at synchronize time, giving the resource an immediate value without waiting for an event.

As part of this work, the event mapper's suppress contract was formally defined and correctly implemented — scripts may return {} to intentionally skip an update (e.g., for non-state-change event sub-types), which is now handled as a clean no-op rather than an error. Returning {output: null} is treated as a script error, distinguishing a bug where the script intended to publish but produced null from a deliberate suppress.

Key design decisions:
- seedFrom is a companion to mapper.event, not a replacement for mapper.read; read+seedFrom is rejected by the parser as a configuration error
- seedFrom reuses the mapper.read script interface (sbmdReadArgs context variable) so the same script shape applies to both
- seedFrom bindings are stored in a separate resourceSeedFromBindings map and are never added to readableAttributeLookup, preventing the attribute from being treated as a subscribable read after the initial seed
- At commission time, DoRegisterResources reads the seed value via ReadSeedValueFromAttribute and passes it directly to createDeviceResource/createEndpointResource, so DEVICE_ADDED carries the correct initial value and clients never observe a null resource at commission
- At synchronize time, DoSynchronizeDevice calls SeedInitialResourceValues, which iterates all resources with seedFromAttribute set and calls SeedResourceFromAttribute (which delegates to ReadSeedValueFromAttribute + updateResource) for each non-skipped resource
- MapEvent has a tri-state contract: false = script error; true + empty outValue = suppress; true + non-empty outValue = publish. A non-object return (null, undefined, primitive) is always a script error, guarded by an object type check that matches the MapWrite/MapExecute pattern. An object with output: null is also a script error.

Implementation:
- SbmdSpec: added seedFromAttribute (std::optional<SbmdAttribute>) and seedFromScript (string) to SbmdMapper; presence of seedFrom is indicated by seedFromAttribute.has_value()
- SbmdParser: parse seedFrom block; reject event alias in seedFrom.alias, missing script, read+seedFrom conflict, and seedFrom-without-event; stamp resource IDs onto seedFromAttribute in SetMapperIds
- sbmd-spec-schema.json: added seedFromMapper definition and seedFrom as optional mapper property
- MatterDevice: added BindResourceSeedFromInfo (populates resourceSeedFromBindings only), ReadSeedValueFromAttribute (cache read + script mapping, returns std::optional<std::string>), and SeedResourceFromAttribute (calls ReadSeedValueFromAttribute + updateResource)
- SpecBasedMatterDeviceDriver: overrides DoRegisterResources to read seed values via ReadSeedValueFromAttribute and pass them as initial values to createDeviceResource/createEndpointResource at commission time; overrides DoSynchronizeDevice to call SeedInitialResourceValues at reconnect time
- SbmdScriptImpl (MapEvent): added object type guard before suppress check; {output: null} returns false with an error log; suppress path ({} or missing output key) clears outValue and returns true rather than calling ExtractScriptOutputAsString; both quickjs and mquickjs backends implement the same tri-state contract
- door-lock.sbmd: locked resource changed from mapper.read to mapper.event (LockOperation, cluster 0x0101, event 0x0002) + mapper.seedFrom (LockState attribute, 0x0000); event script correctly uses sbmdEventArgs; lockOperation added as event alias and as second prerequisite on the locked resource
- DoorLockDevice.js: handleLock and handleUnlock emit LockOperation events in addition to updating lockState; a DoorLockServerWithEvents subclass overrides lockDoor/unlockDoor on the Matter command path to also emit LockOperation events, so Barton-initiated commands update the locked resource
- VirtualDevice.js: added goOffline/comeOnline sideband ops — goOffline calls initiateForceClose() on all active Matter sessions without sending any Matter messages (no SessionClose, no StatusReport), leaving the ServerNode running so mDNS advertisement and the fabric state remain intact; handleComeOnline in the base class is a no-op and DoorLockDevice.js overrides it to update the LockState attribute before Barton reconnects via a new CASE session
- DeviceDriver: added metadataUpdated(driver, device, key, value) callback (parallel to commFailTimeoutSecsChanged), called from setMetadata() in deviceService.c after each persisted metadata change; MatterDeviceDriver sets this callback to handle matterLivenessTimeoutOverrideMs
- DeviceDataCache: OverrideLiveness(uint32_t ms) calls readClient->OverrideLivenessTimeout(Milliseconds32(ms)) via RunOnMatterSync; OnSubscriptionEstablished resets mLivenessTimeoutOverride to Clock::kZero after each successful subscription, preventing rapid-resubscription loops
- deviceService.c (setMetadata): when COMMON_DEVICE_METADATA_COMMFAIL_OVERRIDE_SECS is persisted, calls deviceServiceCommFailSetDeviceTimeoutSecs immediately to reprogram the running per-device watchdog timer without requiring a device reconnect or service restart
- deviceService.c (deviceServiceStart): reads COMM_FAIL_MONITOR_INTERVAL_SECS_PROP after deviceServiceCommFailInit() and calls deviceCommunicationWatchdogSetMonitorInterval() when non-zero (one-shot for values already persisted to the DB); wires an OnCommFailMonitorIntervalPropertyChanged signal handler on BCorePropertyProvider so that writes made after startup also take effect immediately
- deviceCommunicationWatchdogSetMonitorInterval: now broadcasts controlCond after updating the interval, waking the watchdog thread immediately rather than waiting for the current sleep to expire

Tests:
- sbmdParserTest: updated test_sbmdParserDoorLockFile to assert seedFromAttribute.has_value()==true and hasRead==false; added 5 new seedFrom tests covering valid spec, seedFrom-without-event, missing script, read+seedFrom conflict, and event-alias-rejected
- SbmdScriptTest: added MapAttributeReadUint8ToBoolean covering locked (1), unlocked (2), and not-fully-locked (0) states against the door-lock seed script; added 14 MapEvent tests covering the full tri-state contract including all error paths, suppress path, and sbmdEventArgs field exposure
- door_lock_test.py: added test_locked_resource_seeded_on_commission, test_locked_resource_updated_by_event, and test_locked_resource_seeded_on_synchronize; the synchronize test uses only the standard b_core_client_write_metadata API — commFailOverrideSeconds=1 shortens the per-device watchdog timeout, matterLivenessTimeoutOverrideMs=1 collapses the ReadClient liveness window to trigger immediate resubscription, and barton.commFail.monitorIntervalSecs=1 is set inline in the test body after commission (via BCorePropertyProvider) so the watchdog checks every second

Refs: BARTON-355


core/deviceDrivers/matter/sbmd/SbmdScript.h
```diff
@@ -295,10 +295,23 @@ namespace barton
          *     "output": <Barton string representation of the event data>
          * }
          *
+         * If the script omits the "output" key but returns a plain object (e.g., returns {}),
+         * the event is intentionally suppressed: MapEvent returns true with outValue left empty.
+         * The caller MUST check outValue.empty() and skip updateResource in that case.
+         * This is useful when an event type carries multiple operation sub-types, only some of
+         * which represent a resource state change. For example, a LockOperation event may carry
+         * a lock, unlock, or door-sense operation; a script can return {} for sub-types it does
+         * not need to propagate, avoiding spurious resource updates.
```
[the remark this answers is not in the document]
rchowdcmcsa: There is no scenario (so far) where the empty-string case is a valid resource value in Barton. No event mapper would ever have a reason to publish an empty string as a legitimate value. You are technically correct in your concern, but practically speaking I don't think the churn for closing this gap is worth addressing at this time.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
```diff
@@ -986,7 +992,45 @@ bool SbmdScriptImpl::MapEvent(const SbmdEvent &eventInfo,
...
+    {
+        icDebug("Event mapper returned no output (suppressed) for cluster 0x%X, event 0x%X",
+                eventInfo.clusterId,
+                eventInfo.eventId);
+        outValue.clear();
+
+        return true;
+    }
+
     return ExtractScriptOutputAsString(outJson, outValue);
```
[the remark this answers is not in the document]
rchowdcmcsa: Refactoring ExtractScriptOutputAsString() to convert the already-fetched outputVal introduces the potential for user error imo (every call to ExtractScriptOutputAsString() would have to satisfy a precondition that the outputVal was already fetched from outJson and validated -- and that's a lot of code duplication in its own right, unless the fetch/validation is made into its own helper, but then you'd just have to call two helpers everywhere).

MapEvent() pre-fetches "output" to distinguish suppress from publish before any conversion happens. Since we already have the value in hand, I'm going to change MapEvent() to convert it directly in-line to avoid the redundant property lookup that calling ExtractScriptOutputAsString() would introduce. The code duplication introduced in this one spot by doing so is an acceptable trade-off in my mind. The helper remains appropriate for the other mappers where no pre-fetch is needed.

testing/test/door_lock_test.py
tleacmcsa, line 1: What about the synchronize scenario (where barton starts up with an already commissioned lock and updates for a missed state change?

core/deviceDrivers/matter/sbmd/SbmdSpec.h
```diff
@@ -164,6 +164,13 @@ namespace barton
         // Event mapping - for handling Matter events that update the resource
         std::optional<SbmdEvent> event;
         std::string eventScript;
+
+        // SeedFrom mapping - for seeding initial resource value from the attribute cache
+        // at configure and synchronize time. Only valid alongside an event mapper.
+        // YAML key: seedFrom.
+        bool hasInitialRead = false;
```
tleacmcsa: Do we really need this "has" boolean given that optional is used below? Similar to the event stuff above.

core/deviceDrivers/matter/sbmd/SbmdParser.cpp
```diff
@@ -686,6 +709,43 @@ bool SbmdParser::ParseMapper(const YAML::Node &node, SbmdMapper &mapper, const s
...
+
+        if (!seedFromNode["script"] || seedFromNode["script"].as<std::string>().empty())
+        {
+            icError("seedFrom mapper must have a non-empty 'script'");
+            return false;
+        }
+
+        mapper.hasInitialRead = true;
+        mapper.initialReadAttribute = alias->attribute;
+        mapper.initialReadScript = seedFromNode["script"].as<std::string>();
```
tleacmcsa: The naming discrepancy between "seedFrom" and "initialRead" is a bit confusing.  Can we standardize?  I'm not sure which is best.  "seed" to me implies a one-time only thing (during commissioning) and it covers that but also synchronize.  Similar with "initialRead" I suppose.  I don't feel strongly on what is used, but they probably should be the same.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
```diff
@@ -986,8 +993,62 @@ bool SbmdScriptImpl::MapEvent(const SbmdEvent &eventInfo,
...
+    {
+        icError("Event mapper script returned a non-object for cluster 0x%X, event 0x%X",
+                eventInfo.clusterId,
+                eventInfo.eventId);
+
+        return false;
+    }
+
+    // Check whether the script returned an 'output' field.
+    // A missing 'output' on an object return is a valid "suppress" signal — the script ran
```
tleacmcsa, line 1010: Note: BARTON-377 formalizes this script interface

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
```diff
@@ -986,8 +993,62 @@ bool SbmdScriptImpl::MapEvent(const SbmdEvent &eventInfo,
...
+
+    JSCStringBuf buf;
+    const char *resultStr = JS_ToCString(ctx, outputVal, &buf);
+
+    if (!resultStr)
+    {
+        icError("Failed to convert event output to string for cluster 0x%X, event 0x%X: %s",
+                eventInfo.clusterId,
+                eventInfo.eventId,
+                GetExceptionString(ctx).c_str());
```
tleacmcsa, line 1051: if JS_ToCString returns null, are we guaranteed to have an exception?

core/deviceDrivers/matter/sbmd/SbmdScript.h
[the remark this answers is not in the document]
rchowdcmcsa: Would like thoughts on this though @cleithner-comcast @tleacmcsa @kfundecmcsa

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -154,6 +154,16 @@ bool SpecBasedMatterDeviceDriver::AddDevice(std::unique_ptr<MatterDevice> device
                 return false;
             }
         }
+
+        if (sbmdResource.mapper.hasInitialRead && sbmdResource.mapper.initialReadAttribute.has_value())
```
tleacmcsa: hasInitialRead and has_value are redundant (and risky should they ever get out of sync). We should probably just stick with the optional / has_value and ditch hasInitialRead bool.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -383,6 +400,26 @@ bool SpecBasedMatterDeviceDriver::DoRegisterResources(icDevice *device)
     return result;
 }
 
+bool SpecBasedMatterDeviceDriver::DevicePersisted(icDevice *device)
+{
+    // At this point the device has been added to the database and DEVICE_ADDED has been emitted,
```
tleacmcsa: so the device will be first added with incomplete/missing resources, then updated? Ideally it is created fully populated first. Clients will probably send data to the cloud on the added event since they would not know to expect updated events later.  so it can be inefficient on whatever they might do (send to cloud, etc.).

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -495,6 +532,60 @@ std::string SpecBasedMatterDeviceDriver::MakeResourceKey(const SbmdResource &res
...
+
+    auto device = GetDevice(deviceId);
+
+    if (device == nullptr)
+    {
+        icError("Device %s not found for seedFrom seeding", deviceId.c_str());
+        return;
+    }
+
+    auto skipIt = skippedOptionalResources.find(deviceId);
```
tleacmcsa: This is now the second place that iterates and accounts for skipped optional resources.  Can you add some smart iteration helper methods instead of duplicating the pattern?  Each time its done we run the risk of doing it differently.

core/deviceDrivers/matter/MatterDevice.cpp
```diff
@@ -634,6 +642,148 @@ bool MatterDevice::BindResourceEventInfo(const char *uri,
...
+    {
+        icDebug("seedFrom attribute not in cache for URI: %s (cluster 0x%x, attribute 0x%x): %s",
+                uri,
+                static_cast<uint32_t>(binding.attributePath.mClusterId),
+                static_cast<uint32_t>(binding.attributePath.mAttributeId),
+                err.AsString());
+        return;
+    }
+
+    if (!script)
```
tleacmcsa: It's best to have prerequisites as high as possible to avoid unnecessary work.

openspec/specs/sbmd-system/spec.md
```diff
@@ -134,6 +134,21 @@ A resource's mapper MAY contain an `event` section with an `alias` (a string nam
...
+- **THEN** the resource SHALL NOT be updated and no error SHALL be logged — this is the standard mechanism for ignoring non-state-change events
+
+### Requirement: seedFrom mapper type
+A resource mapper SHALL support a `seedFrom` section (in addition to the existing `read`, `write`, `execute`, and `event` sections) for populating initial resource values from the attribute cache when the resource's ongoing updates are driven by a `mapper.event`. The `seedFrom` mapper SHALL NOT be used as a substitute for `mapper.read` — if ongoing attribute subscription updates are desired, `mapper.read` remains the correct choice. The `seedFrom` mapper SHALL use the `sbmdReadArgs` script interface (identical to `mapper.read`).
+
+#### Scenario: Event-driven resource has initial value at commission
+- **WHEN** a resource declares `mapper.event` for ongoing updates and `mapper.seedFrom` pointing to a corresponding attribute alias
+- **THEN** the resource SHALL have a non-null value immediately after device commissioning without waiting for the first event to fire
+
+#### Scenario: Event-driven resource has initial value after hub restart
```
tleacmcsa: we probably shouldnt use 'hub' terminology and instead refer to Barton.  Minor.

tleacmcsa (changes requested):

core/deviceDrivers/matter/sbmd/SbmdScript.h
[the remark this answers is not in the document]
tleacmcsa: We can clean this up as part of BARTON-377

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
rchowdcmcsa: We are not. GetExceptionString() handles the case where no exception is thrown by just returning `"unknown error"`, but we can change the log message anyway. I don't feel strongly about it.

testing/test/door_lock_test.py
rchowdcmcsa: There are two ways to go about it.

One is to test it by having the virtual device restart and then reconnect. This would require changing `VirtualDevice.js` to accept a `--storage <path>` CLI argument and use it as `path.root` instead of `fs.mkdtempSync()`, so the virtual device retains its Matter fabric credentials (NOC) across stop/start. Also we'd have to set a short commfail timeout via the `COMMON_DEVICE_METADATA_COMMFAIL_OVERRIDE_SECS` device metadata, but that's easy enough. I haven't tried implementing this yet.

The second way (as you described) is by testing the path where Barton starts and has an already commissioned lock that it needs to check. I tried implementing a test like this real quick where I restart Barton mid-test, and it didn't work because stopping and restarting Barton during the same subprocess caused a VerifyOrDie crash in the InetLayer. This is because we have Matter implemented as a C++ singleton that cannot reinitialize within a single process. (`matterSubsystem.cpp:213` makes note of this tech debt: "FIXME: Matter shouldn't be a singleton with Init() but a properly constructed object," but anyway.)

There's some stuff in `conftest.py` that looks like it can help with this. I talked to Claude and the flow would look something like:

```
outer pytest
└── child process 1  (spawned by pytest_runtest_protocol)
    ├── Phase 1: commission, unlock via sideband, stop BCore client
    │            (device DB now on disk at ~/.brtn-ds)
    └── spawn child process 2  (via subprocess.run, just a Python script)
        └── Phase 2: start fresh BCore client against same ~/.brtn-ds
                     → deviceInitializationTask → DoSynchronizeDevice
                     → SeedInitialResourceValues reads LockState = unlocked
                     → assert locked resource = "false"
```

Laying this out here to ask for thoughts, since I don't really know how best to proceed. Maybe both?

tleacmcsa: The second path seems hard.  The other way is to come out of communication failure.  I'm not sure if thats what you were describing for the first option, but its much easier than that. Fake it... artificially put it in comm fail then cause it to come out.  That should hit MatterDeviceDriver.cpp's communicationRestored, which does the synchronizeDevice call.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
tleacmcsa: as long as it doesnt crash when there is no exception, i think its ok.

testing/test/door_lock_test.py
cleithner-comcast: Tangentially related, barton should not be crashing if we stop/start it in-process. That's a major lifecycle problem in my mind. If that's occuring, can you file a bug with some diagnostics? I'd like to get it scheduled.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: Moved logic to DoRegisterResources

testing/test/door_lock_test.py
rchowdcmcsa: Wrote a test.
Note that I added test scaffolding to MatterDeviceDriver and DeviceDataCache to allow for overriding ReadClient's liveness timeout before resubscription is triggered, else the test would have to wait like 15s for the device to come out of commfail.

Here's that bug:
https://ccp.sys.comcast.net/browse/BARTON-378

api/c/public/private/bartonMatterTestHelpers.h
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Intended solely for integration tests that need to exercise the
+ * communicationRestored → synchronizeDevice path without waiting for the
+ * full negotiated liveness window (which can be tens of seconds).
+ *
+ * MUST NOT be called from production code or any Barton client application.
+ *
+ * @param uuid The device UUID as registered in the device service.
+ */
+void bartonMatterDeviceForceResubscription(const char *uuid);
```
cleithner-comcast: This feels too direct for an integration test. I would expect a test to tamper with properties to speed up operations and use the mock device to force offline/online.

core/deviceDrivers/matter/sbmd/SbmdParser.cpp
tleacmcsa: this conversation should not have been marked resolved yet.  we should discuss when to do that as a team so everyone is on the same page.  I pointed out a concern without a clear "I would like it changed to X" message, so the solution should have been reviewable.  If resolved, it is very hard for the reviewer to see how you addressed a non-trivial comment.

testing/test/door_lock_test.py
```diff
@@ -119,3 +119,116 @@ def test_sideband_lock_triggers_barton_update(
...
+    # The ServerNode stays running so Barton can reconnect once asked to.
+    matter_door_lock.sideband.send("goOffline")
+
+    # Force comm-fail immediately — device is genuinely unreachable at this point.
+    libbarton = ctypes.CDLL(None)
+    libbarton.deviceCommunicationWatchdogForceDeviceInCommFail.argtypes = [
+        ctypes.c_char_p
+    ]
+    libbarton.deviceCommunicationWatchdogForceDeviceInCommFail.restype = None
+    libbarton.deviceCommunicationWatchdogForceDeviceInCommFail(lock.props.uuid.encode())
```
cleithner-comcast: This shouldn't be how we go about doing commfail. In ZITH we would speed up detection to be on the time scale of tests. That was sufficient for zigbee devices. For matter, I figure we should be able to do the same with subscriptions.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
```diff
@@ -1620,3 +1621,64 @@ static bool getDeviceClassVersion(void *ctx, const char *deviceClass, uint8_t *v
...
+// The functions below (MatterDeviceDriver::ForceResubscription and
+// bartonMatterDeviceForceResubscription) exist solely to support integration
+// tests that need to exercise the communicationRestored → synchronizeDevice
+// path quickly, without waiting for the full negotiated liveness window.
+//
+// MUST NOT be called from production code or any Barton client application.
+// The public test API is declared in api/c/public/private/bartonMatterTestHelpers.h.
+// =============================================================================
+
+void MatterDeviceDriver::ForceResubscription(const std::string &deviceUuid)
```
cleithner-comcast: I think I don't want this code. I hope we can do the test another way but if not I'd rather not have this or the test. Philosophically, I'm a bit of purist when it comes to to integration testing. Introducing code into the black box you have under test just so you can manipulate the instruction execution but only during that test is very circular. Configuration (like properties, or meddling with local files) is one thing, but introducing instructions is a step up that I don't like very much.

cleithner-comcast (commented):
I am not requesting change just yet, but I am opening this up for others to comment.

core/deviceDrivers/matter/MatterDeviceDriver.cpp
rchowdcmcsa: Yeah, that makes sense.

To clarify some stuff, the reason I did it this way is because the Matter SDK's liveness timeout is derived from MRP retransmission params and is, by default, around 16-17 seconds. (There are some other things that go into calculating the liveness timeout that we could manipulate on the virtual device end, but not enough to get the timeout below 15 or so seconds.) The only way to bypass this computation entirely is through the SDK's OverrideLivenessTimeout API. Our watchdog sits downstream of any Matter messages sent over the Matter session (see SessionMessageHandler). The comm restore must be triggered via a CASE reconnect producing an authenticated message. And a CASE reconnect only happens after ReadClient terminates the subscription and DefaultResubscribePolicy schedules one. And ReadClient only does that when the liveness timer fires.

I considered faking the commfail in some way without actually killing the Matter sessions and then faking the comm restore to get around the aforementioned liveness timeout stuff, but then the test path no longer reliably tests the communicationRestored -> synchronizeDevice flow, because any sideband state change on the device is delivered to Barton over the still-open subscription. So the test would not be able to tell if the resource was populated via seedFrom in the synchronzeDevice path or via a regular event update. Maybe there's a correct way to do this still, but I didn't land on it.

I would be OK with punting on this integration test, but also open to other ideas on how we might implement it.

cleithner-comcast: That's a great explanation, thanks for learning me. Would it makes sense to introduce a runtime property interface for OverrideLivenessTimeout, setting it to a very small test value? As in, would that alleviate the deeply penetrating "test only" private but public API?

rchowdcmcsa: I gave it a whirl. Let me know if this is any better.

tleacmcsa: Since the point is to test the synchronize path, why not trigger the startup/load instead of the commfail?

rchowdcmcsa: Like, Barton startup? I tried that before but the "Matter singleton cannot re-init within a single process" problem surfaced. There's some stuff in `conftest.py` that could probably help us get around it by utilizing a two-subprocess handoff, but we deemed it too difficult.

tleacmcsa: That is re-init.  What if the config was put down first, then we start up?  That would be a single init, right?

rchowdcmcsa: Could work. I'll explore that path and get back to ya.

rchowdcmcsa: @tleacmcsa Putting down the config requires persisting the Barton Matter KVS and the matterjs device's fabric storage, and the on-disk format of that data is owned by the Matter SDK. Any SDK version bump that changes the storage format would silently break the test or require regenerating this config (either from a snapshot from a real commissioning run, or by just making it up somehow (not sure how that'd go)), so the test would require active maintenance down the line. Maybe there's a better way to go about doing what you suggested, but that's what I've found.

The other way is how I described it before, with two subprocesses within one test (something like the diagram below) but that's even more complicated:

```
outer pytest
└── child process 1  (spawned by pytest_runtest_protocol)
    ├── Phase 1: commission, unlock via sideband, stop BCore client
    │            (device DB now on disk at ~/.brtn-ds)
    └── spawn child process 2  (via subprocess.run, just a Python script)
        └── Phase 2: start fresh BCore client against same ~/.brtn-ds
                     → deviceInitializationTask → DoSynchronizeDevice
                     → SeedInitialResourceValues reads LockState = unlocked
                     → assert locked resource = "false"
```

I dunno. I feel like, if we're gonna have this test at all, the commfail way is just easier than these (once I fix the bug that's causing it to fail right now).

testing/test/door_lock_test.py
```diff
@@ -38,6 +41,28 @@
 pytestmark = pytest.mark.requires_matterjs
 
 
+@pytest.fixture
+def fast_commfail_environment():
```
cleithner-comcast: This shouldn't be defined in test file scope, test infrastructure should be separate (we have environments).

With that said, I don't like the idea of having a bunch of different environments for small things (I can imagine most tests want or don't care if comm fail is configured to be fast). I think we should either make fast comm fail the default or make it something you configure for the fixture by the test (see @pytest.mark.parametrize)

testing/test/door_lock_test.py
```diff
@@ -119,3 +144,132 @@ def test_sideband_lock_triggers_barton_update(
...
+
+    Both properties are set via the standard b_core_client_write_metadata API.
+    """
+
+    lock = _commission_door_lock(fast_commfail_environment, matter_door_lock)
+    client = fast_commfail_environment.get_client()
+
+    # Shorten the comm-fail watchdog so it fires in ~3 s instead of ~56 min.
+    metadata_base = f"/{lock.props.uuid}/m"
+    client.write_metadata(f"{metadata_base}/commFailOverrideSeconds", "3")
```
cleithner-comcast: I don't understand setting metadata on the device for comm fail override. Is that an existing interface? I'd expect a generic "fast timer" prop for comm fail watchdog that can be set in the environment before we run BartonClient.start().

Also 3 seconds is a very long time for a test.

testing/test/door_lock_test.py
```diff
@@ -119,3 +144,132 @@ def test_sideband_lock_triggers_barton_update(
...
+
+    commfail_queue = resource_update_listener(client, "communicationFailure")
+
+    # Abruptly kill all Matter sessions without sending any Matter messages.
+    # The ServerNode stays running so Barton can reconnect once asked to.
+    matter_door_lock.sideband.send("goOffline")
+
+    # Wait for Barton to detect comm-fail via the watchdog (~4 s nominally, but
+    # CI runners can be slow so a generous timeout is used).
+    wait_for_resource_value(commfail_queue, "true", timeout=20)
```
cleithner-comcast: 20 seconds is a long timeout, I would err on like 5.

testing/test/door_lock_test.py
```diff
@@ -119,3 +144,132 @@ def test_sideband_lock_triggers_barton_update(
...
+    matter_door_lock.sideband.send("comeOnline", {"lockState": "unlocked"})
+
+    # The ReadClient's liveness timer still has ~14 s remaining at this point.
+    # Setting matterLivenessTimeoutOverrideMs=1 causes MatterDeviceDriver to
+    # apply ReadClient::OverrideLivenessTimeout(1ms) via ScheduleLambda, so
+    # the liveness fires immediately, triggering DefaultResubscribePolicy to
+    # open a new CASE session.  Since the device is now online, CASE succeeds,
+    # the priming report delivers LockState=Unlocked, and synchronizeDevice
+    # seeds locked="false".
+    client.write_metadata(f"{metadata_base}/matterLivenessTimeoutOverrideMs", "1")
```
cleithner-comcast, line 214: Again, I'd expect a blanket property for all matter devices when it comes to liveness timeout override. I don't think it makes sense to configure these after we have started barton up on a per-device  basis

cleithner-comcast (changes requested):

testing/test/door_lock_test.py
rchowdcmcsa: I took the custom environment out; there's no need for it now after I made a few fixes.

testing/test/door_lock_test.py
rchowdcmcsa: Changed to 5.

testing/test/door_lock_test.py
cleithner-comcast: Posting some comments from Raiyan from slack:

"The comm fail override metadata key existed before, but Matter devices didn't do anything with it (they don't participate in DDL logic), hence the additions [here](https://github.com/rdkcentral/BartonCore/pull/211/changes#diff-438278c4732d6c06aed604ec0f769f5b2ee829c4ada037c87666ce7a8cc89499), specifically wiring deviceServiceCommFailSetDeviceTimeoutSecs() so that writing commFailOverrideSecs after commissioning reprograms the running watchdog timer. commFailOverrideSecs is inherently a per-device metadata that determines how long we wait before declaring that device to be in commfail, and the metadata write path requires a UUID (obtained after commissioning, hence why it comes after start_client). I tried exploring another way to do this but couldn't land on it (you might recall that I called "ForceCommFail()" or whatever it was called in a previous iteration, that's the only other thing I can think of).
The "fast timer" prop you refer to is essentially what the newly added monitorIntervalSecs property is, if I understood you correctly -- both are necessary, one to control the commfail timeout, the other to control how quickly the watchdog wakes up to detect the commfail."

testing/test/door_lock_test.py
cleithner-comcast: Posting some comments from Raiyan from slack:

"matterLivenessTimeoutOverrideMs could maybe be a blanket property set before start_client but it would be tricky. Right now, it basically functions as a way to tell an existing ReadClient subscription to use a short liveness timeout so that resubscription after commfail can happen quickly. If it was a blanket property, then it would have to broadcast the liveness timeout override to all active running subscriptions, which is strange to me. And setting a property pre-startup is tricky because the way it works right now is we use the liveness timer override as a way to basically say "okay, resubscribe to the doorlock now." Maybe it could work as a pre-startup property but I dunno. This is admittedly awkward because this is a hook I had to add to make the test work (even though a hook that says "force a CASE resubscription to this specific device right now" might have its uses in other contexts)."

cleithner-comcast (dismissed):
Needs to be updated with main but I'm generally good with this as is. Leaving some comments open in case anyone else wants to chime in. Raiyan, feel freem to resolve any conversations I started if/when you're ready to move the change forward.

tleacmcsa (dismissed):
I agree with Christian's message.  We can get some experience with it like this and see if we want to tweak going forward.  If there are any actions out of the comments that we want to do later, lets get some tickets for them, but I think we are basically good.

kfundecmcsa (dismissed):

kfundecmcsa (approved):

tleacmcsa (approved):
cant tell what changed, but i was told only rebase.  trusting! :-)

cleithner-comcast (approved):

PR #214, opened 2026-05-01
fix(config): switch zigbee region lookup to cpe.zigbee.region_code

NaeemKK:
Replace legacy country_code key with cpe.zigbee.region_code and update zigbee subsystem lookup to the new property.

Refs: XHCPE-2544


core/src/devicePrivateProperties.h
```diff
@@ -50,7 +50,7 @@
 #define TEST_FASTTIMERS_PROP                                  "cpe.tests.fastTimers"
 #define CPE_ZIGBEE_CHANNEL_CHANGE_ENABLED_KEY                 "cpe.zigbee.channelChange.enabled"
 #define CPE_ZIGBEE_CHANNEL_CHANGE_MAX_REJOIN_WAITTIME_MINUTES "cpe.zigbee.channelChange.maxWaitTimeMin"
-#define CPE_REGION_CODE_PROPERTY_NAME                         "country_code"
+#define CPE_ZIGBEE_REGION_CODE_PROPERTY_NAME                  "cpe.zigbee.region_code"
```
tleacmcsa: since we are changing it, might as well make the name conform to the pattern (no _PROPERTY_NAME at the end).

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -792,8 +792,9 @@ static bool initializeNetwork(void)
     }
 
     uint64_t localEui64 = loadLocalEui64();
-    scoped_generic char *region = b_core_property_provider_get_property_as_string(
-        propertyProvider, CPE_REGION_CODE_PROPERTY_NAME, "US");
+
+    scoped_generic char *region =
+        b_core_property_provider_get_property_as_string(propertyProvider, CPE_ZIGBEE_REGION_CODE, "US");
```
tleacmcsa: As I mentioned in the other review, I'm thinking that we dont provide a default of "US" here.  If we don't get a property dont send it in the init and let ZigbeeCore decide what to do since it knows default regions, etc.

NaeemKK: Done

core/src/devicePrivateProperties.h
NaeemKK: Done

tleacmcsa:
looks good.  update the branch and take out of draft and ill be good with it.

core/src/subsystems/zigbee/zigbeeSubsystem.c
```diff
@@ -792,8 +792,9 @@ static bool initializeNetwork(void)
     }
 
     uint64_t localEui64 = loadLocalEui64();
-    scoped_generic char *region = b_core_property_provider_get_property_as_string(
-        propertyProvider, CPE_REGION_CODE_PROPERTY_NAME, "US");
+
+    scoped_generic char *region =
+        b_core_property_provider_get_property_as_string(propertyProvider, CPE_ZIGBEE_REGION_CODE, NULL);
```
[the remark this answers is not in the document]
NaeemKK: resolved

tleacmcsa (approved):

core/src/subsystems/zigbee/zigbeeSubsystem.c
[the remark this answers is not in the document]
NaeemKK: Yes, the intention is to avoid sending any value when the region is not defined, and let the client handle it appropriately. If the region is available, we’ll include it otherwise, we’ll leave it as NULL.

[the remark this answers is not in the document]
NaeemKK: I think it’s pretty straightforward. if it can retrieve a value using the property service library, it will set it; otherwise, it will just leave it as NULL.

cleithner-comcast (approved):

cleithner-comcast:
FYI the org requires all comments to be marked resolved before we can merge, including copilot.

Also don't forget to update with main.

kfundecmcsa (approved):

PR #220, opened 2026-05-22
build(docker): Add optional debuginfod volume

cleithner-comcast:
This commit adds an optional debuginfod volume compose file. If there is a DEBUGINFOD_CACHE_PATH environment variable set, it will use that. Else, it will use the default $HOME/.cache/debuginfod_client path. This will allow people to persist a cache across BartonCore containers.


cleithner-comcast:
Closing for https://github.com/rdkcentral/BartonCore/pull/248 for now.

PR #221, opened 2026-05-27
feat(sbmd): replace bool+out-param mapper API with ScriptResult

rchowdcmcsa:
Replaces the `bool` return / output-parameter pattern across all SBMD script mapper methods with a typed `ScriptResult` value. This gives callers a first-class way to distinguish between a successful value update, an intentional skip of the resource update, and a script error — without relying on output-parameter mutation or sentinel empty strings.

Changes:

New `ScriptResult` type (`ScriptResult.h` / `ScriptResult.cpp`):
- Encapsulates three outcomes: error, skip-resource-update, or an operation payload.
- The payload is a `std::variant<ResourceUpdate, ScriptWriteResult>`, covering both attribute-read/event results and write/execute results.
- Factory helpers: `MakeError`, `MakeSkipResourceUpdate`, `MakeResourceUpdate`, `MakeWriteResult`.
- `HasOperation()`, `IsError()`, `SkipsResourceUpdate()`, `Operation()` accessors.
- `DecodeTlvBase64` utility (moved from `SbmdScriptImpl`) with an explicit bounds check before the `uint16_t` cast.

`SbmdScript` API (`SbmdScript.h`)

All five virtual mapper methods now return `ScriptResult` instead of `bool`:
- `MapAttributeRead`
- `MapCommandExecuteResponse`
- `MapWrite`
- `MapExecute`
- `MapEvent`

Script return value key (all `.sbmd` specs, `sbmd-utils.js`, `sbmd-script.d.ts`)

Scripts now return `{ value: ... }` instead of `{ output: ... }`. Schema version bumped to `3.0`. A new `seedFrom` field allows event-driven resources to seed their initial value from a readable attribute.

Implementations (`quickjs/SbmdScriptImpl.cpp`, `mquickjs/SbmdScriptImpl.cpp`):
- Both engines updated to construct and return `ScriptResult` values.
- Shared per-type property-extraction lambdas (`getPropertyUint` / `getPropertyStr`) replace duplicated lambda bodies.
- Clarifying comment added above `JS_GetException` explaining ownership transfer semantics.

`MatterDevice` — call sites updated to consume `ScriptResult`.

SBMD schema — new `v3/` schema directory; `CHANGELOG.md` updated. Also fixes a pre-existing omission where `seedFromMapper` was missing from the v2.1 schema.

Docs — `SBMD.md` updated to document `value`, `suppress`, `seedFrom`, and the new schema version.

Test changes:
- `ScriptResultTest.cpp` (new): unit tests for all `ScriptResult` factory helpers and the `DecodeTlvBase64` / `MakeFromJson` paths.
- `MatterDeviceTest.cpp` (new): extracted from `MatterDeviceEndpointMapTest` to give `MatterDevice` its own focused test binary.
- `MatterDeviceTestHelpers.h` (new): shared test header (`TestableMatterDevice`, `MockSbmdScript`, `ChipPlatformEnvironment`) used by both test binaries to eliminate drift.
- `SbmdScriptTest.cpp`, `sbmdParserTest.cpp` — updated for the new API and `value` return key.
- `CMakeLists.txt` — registers `testMatterDevice` and `testScriptResult` as independent test targets.

Refs: BARTON-377, BARTON-366


core/test/src/MatterDeviceTest.cpp
```diff
@@ -0,0 +1,198 @@
...
+        uint8_t tlvBuffer[kTlvBufferSize];
+        uint32_t written = buildMinimalTlvBuffer(tlvBuffer, kTlvBufferSize);
+
+        chip::TLV::TLVReader reader;
+        reader.Init(tlvBuffer, written);
+        reader.Next();
+
+        chip::app::EventHeader header = MakeTestEventHeader();
+        device->GetCacheCallback()->OnEventData(header, &reader, nullptr);
+    }
```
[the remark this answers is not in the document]
rchowdcmcsa: The structural check is fine as-is. OnEventData() has two independent guards that prevent updateResource() from being called when the result is suppressed. In my opinion, this is sufficient enough that the churn of wrapping updateResource behind an injectable interface isn't worth it.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdScriptImpl.cpp
```diff
@@ -60,34 +62,261 @@ namespace barton
...
             }
 
             // If that fails, try to get the "message" property (for Error objects)
-            if (JS_IsPtr(ex))
+            if (result.empty() && JS_IsPtr(ex))
             {
+                JSGCRef msgVal_ref;
                 JSValue msgVal = JS_GetPropertyStr(ctx, ex, "message");
+                JS_PUSH_VALUE(ctx, msgVal);
+
```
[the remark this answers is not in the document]
rchowdcmcsa: The ex_ref and msgVal_ref variables are not dead code -- they're the required companion structs for the JS_PUSH_VALUE/JS_POP_VALUE macros, which use token-pasting (v ## _ref) to reference them by name.

So JS_PUSH_VALUE(ctx, ex) expands to: `JS_PushGCRef(ctx, &ex_ref); ex_ref.val = ex;`

Both &ex_ref and ex_ref.val = ... are real uses that the compiler recognizes -- no -Wunused-variable warning is generated (confirmed by the current build output). JS_PUSH_VALUE is a void macro with no return value, so there's nothing to assign from it. The code is correct as-is.

core/deviceDrivers/matter/sbmd/SbmdScript.h
```diff
@@ -112,19 +81,16 @@ namespace barton
          *     "attributeType": <attribute type from spec>
          * }
          *
-         * Script output JSON:
-         * {
-         *     "output": <Barton string representation of the attribute value>
-         * }
+         * Script output JSON — one of:
+         * { "value": <string|number|boolean> }  // update resource (non-string coerced to string)
+         * { }  or  { "value": null }            // suppress — no update, no error
```
cleithner-comcast: How does this reconcile with null being a valid resource value? Or is it not anymore?

core/deviceDrivers/matter/sbmd/ScriptResult.cpp
```diff
@@ -0,0 +1,323 @@
...
+        constexpr const char *kKeyWrite = "write";
+        constexpr const char *kKeyError = "error";
+        constexpr const char *kKeyClusterId = "clusterId";
+        constexpr const char *kKeyCommandId = "commandId";
+        constexpr const char *kKeyAttributeId = "attributeId";
+        constexpr const char *kKeyEndpointId = "endpointId";
+        constexpr const char *kKeyTimedInvokeTimeoutMs = "timedInvokeTimeoutMs";
+        constexpr const char *kKeyTlvBase64 = "tlvBase64";
+
+        bool DecodeTlvBase64(const std::string &base64Str,
```
cleithner-comcast, line 57: We should consider migrating the implementation to a central C++ area and have the js runtimes create bindings for them so that we dont have to maintain duplicate code.

core/test/src/MatterDeviceEndpointMapTest.cpp
cleithner-comcast, line 1: Seems like we lost a lot here that's not necessarily relevant.

core/deviceDrivers/matter/sbmd/quickjs/SbmdScriptImpl.cpp
tleacmcsa, line 1: Did you test these changes with the quickjs interpreter too?

core/test/src/MatterDeviceEndpointMapTest.cpp
rchowdcmcsa: I extracted a bunch of helpers from here into "MatterDeviceTestHelpers.h," because the helpers are needed in two test files.

core/deviceDrivers/matter/sbmd/SbmdScript.h
rchowdcmcsa: > Or is it not anymore?

I'm not sure actually. I know the subject has come up several times and each time we've talked about moving away from it, and I think that was in my head when I drew up the design for this.

It would be easy enough to tell Copilot to change this (treat null like a value, not a suppress signal) if we still need to revisit this subject later.

core/deviceDrivers/matter/sbmd/ScriptResult.cpp
rchowdcmcsa: by "implementation," you mean the script result object implementation as a whole, not just the function you highlighted? with the goal to reduce code duplication like in each "extract 'error' key" block and so on?

core/test/src/MatterDeviceEndpointMapTest.cpp
cleithner-comcast: Ah that makes sense.

core/deviceDrivers/matter/sbmd/ScriptResult.cpp
cleithner-comcast: In this case, I was referring to the TLV decode routine since it also exists in sbmdUtils but yes any duplicate logic should be implemented in native code and then bindings should be generated ideally. We haven't yet done this though so if we agree to kick out to avoid scope creep, I'm fine with that. I don't remember if we made a ticket for this previously yet.

rchowdcmcsa: this? seems related at least
https://ccp.sys.comcast.net/browse/BARTON-365

Either way I'm also ok with kicking it out for now.

cleithner-comcast: Yeah, that's the one!

core/deviceDrivers/matter/sbmd/SbmdScript.h
tleacmcsa: lets just create a ticket for "Resolve null resource values" where the task is to figure out what we have now, what we want, and make any changes required.

core/deviceDrivers/matter/sbmd/ScriptResult.cpp
```diff
@@ -0,0 +1,323 @@
...
+extern "C" {
+#include <icLog/logging.h>
+}
+
+namespace barton
+{
+
+    namespace
+    {
+        constexpr const char *kKeyValue = "value";
```
cleithner-comcast: forgot to say I'm not the biggest fan of the k constant prefix. We have strong typing, we don't need the redundancy.

tleacmcsa: agree

core/deviceDrivers/matter/sbmd/specs/thermostat.sbmd
```diff
@@ -126,12 +126,12 @@ endpoints:
             script: |
               var value = SbmdUtils.Tlv.decode(sbmdReadArgs.tlvBase64);
               if (value === null) {
-                return {output: null};
+                return {value: null};
```
tleacmcsa, line 129: isnt this an error?  since this PR adds the error handling, we probably should make sure to return errors where applicable.

core/deviceDrivers/matter/sbmd/ScriptResult.h
```diff
@@ -0,0 +1,193 @@
...
+     */
+    struct ScriptWriteResult
+    {
+        enum class OperationType
+        {
+            Invoke, // Command invocation
+            Write   // Attribute write
+        };
+
+        OperationType type = OperationType::Invoke;
```
tleacmcsa: i prefer to initialize to a value that can indicate unknown/not set/invalid.  If something doesnt set the type, that should be an error instead of assuming invoke.

core/deviceDrivers/matter/sbmd/ScriptResult.h
```diff
@@ -0,0 +1,193 @@
...
+         * Returns true if the error field is set (script reported failure).
+         */
+        bool IsError() const { return error.has_value(); }
+
+        /**
+         * Returns true if neither error nor operation is set.
+         * This is a derived convenience: the script ran successfully but
+         * intentionally produced no action.
+         */
+        bool IsSuppressed() const { return !error.has_value() && !operation.has_value(); }
```
tleacmcsa: I wish we had come up with a better name than suppressed.  A ScriptResult that does not intend to update the resource might be more clear if this was "ShouldUpdateResource" or something.  "Suppress" could be many things (suppress errors, suppress events, etc.).

Maybe another way to think about this concept isnt the action of suppressing, but the decision to not return a new resource value?

core/deviceDrivers/matter/sbmd/ScriptResult.h
```diff
@@ -0,0 +1,193 @@
...
+
+        /**
+         * Returns the error message. Only valid when IsError() is true.
+         */
+        const std::string &ErrorMessage() const { return error.value(); }
+
+        /**
+         * Returns the operation variant. Only valid when HasOperation() is true.
+         */
+        const std::variant<ResourceUpdate, ScriptWriteResult> &Operation() const { return operation.value(); }
```
tleacmcsa, line 136: Feels like this wants to be polymorphic.  As we add operations, would this variant list just keep growing?

core/deviceDrivers/matter/sbmd/quickjs/SbmdScriptImpl.cpp
rchowdcmcsa: Oh, no I did not. How would I go about doing that?

core/deviceDrivers/matter/sbmd/ScriptResult.h
rchowdcmcsa: IsSuppressed() -> IsNoOp()?

core/deviceDrivers/matter/sbmd/specs/thermostat.sbmd
rchowdcmcsa: The attribute is a nullable so I'm not actually sure that it's an error. If a nullable being null is something we should treat as an error, then yeah I can make that change.

core/deviceDrivers/matter/sbmd/ScriptResult.h
rchowdcmcsa: > As we add operations, would this variant list just keep growing?

that was the idea.

I feel like this makes sense? The operations that we support are a fixed enumeration set defined by the json schema, so a fixed list of variants makes more sense to me than a freely extensible base class.

core/deviceDrivers/matter/sbmd/SbmdScript.h
rchowdcmcsa: created
https://ccp.sys.comcast.net/browse/BARTON-389

core/deviceDrivers/matter/sbmd/ScriptResult.h
tleacmcsa: we can live with the "rule of 3" here.  When we add that 3rd operation I bet we will want to switch this.

core/deviceDrivers/matter/sbmd/ScriptResult.h
tleacmcsa: to me "no op" is almost the same thing.  This boolean check has a very specific meaning: does this result intend to update the associated resource or not ?

core/deviceDrivers/matter/sbmd/specs/thermostat.sbmd
tleacmcsa: ok, im fine with that for this attribute.  Please check the other places to ensure we arent missing opportunities to use the new error field you added in this PR.

core/deviceDrivers/matter/sbmd/SbmdScript.h
tleacmcsa: fine to resolve this if @cleithner-comcast is (since he opened the discussion).

core/deviceDrivers/matter/sbmd/quickjs/SbmdScriptImpl.cpp
rchowdcmcsa: tested

tleacmcsa (dismissed):

cleithner-comcast (dismissed):

cleithner-comcast:
The test failure is the known leak sanitizer bug, fixed by https://github.com/rdkcentral/BartonCore/pull/225. You can try re-running until that gets merged as its a race condition or just wait for it to be merged.

cleithner-comcast (approved):

tleacmcsa (approved):

PR #223, opened 2026-05-28
fix: race condition in deviceCommunicationWatchdog shutdown

cleithner-comcast:
The commFailWatchdogThreadProc does not re-check the `running` flag after waking from its timed wait. When deviceCommunicationWatchdogTerm() sets running=false and broadcasts the condition variable, the watchdog thread wakes up but proceeds to iterate the hash table and invoke failedCallback into device drivers that may already be partially shut down by deviceServiceShutdown(). This can corrupt memory and cause a SEGV (detected by AddressSanitizer as use-after-free/double-free) when the hash table is subsequently destroyed.

Fix:
1. Re-check `running` after the condition variable wait returns, so the thread exits immediately when signaled to stop rather than performing one more (unsafe) iteration.
2. NULL out failedCallback/restoredCallback under controlMutex BEFORE broadcasting/joining the thread, so any in-flight callbacks on other paths (PetDevice, ForceDeviceInCommFail) see NULL and skip the call.
3. Guard failedCallback/restoredCallback call sites against NULL to prevent NULL-dereference races during shutdown.

### DIFF
--- a/core/src/deviceCommunicationWatchdog.c
+++ b/core/src/deviceCommunicationWatchdog.c
@@ -247,7 +247,7 @@ void deviceCommunicationWatchdogPetDevice(const char *uuid)
         restoredCb = restoredCallback;
         pthread_mutex_unlock(&controlMutex);
 
-        if (restoredCb != NULL)
+        if (restoredCb)
         {
             restoredCb(uuid);
         }
@@ -310,7 +310,7 @@ void deviceCommunicationWatchdogForceDeviceInCommFail(const char *uuid)
         failedCb = failedCallback;
         pthread_mutex_unlock(&controlMutex);
 
-        if (failedCb != NULL)
+        if (failedCb)
         {
             failedCb(uuid);
         }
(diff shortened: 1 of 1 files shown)


rchowdcmcsa (dismissed):

kfundecmcsa (dismissed):

core/src/deviceCommunicationWatchdog.c
```diff
@@ -450,6 +471,16 @@ static void *commFailWatchdogThreadProc(void *arg)
             incrementalCondTimedWaitMillis(&controlCond, &controlMutex, monitorThreadSleepSeconds);
         }
 
+        // Re-check running after wakeup — if Term() signaled us, exit
+        // immediately rather than iterating devices and calling into
+        // potentially half-torn-down subsystems.
+        if (running == false)
```
tleacmcsa: Some old style stuff here and other places... "== false" and icLogInfo.

tleacmcsa (dismissed):

core/src/deviceCommunicationWatchdog.c
cleithner-comcast: Good call out

tleacmcsa (approved):

rchowdcmcsa (approved):

PR #225, opened 2026-05-29
build(docker): Use LLVM by default

cleithner-comcast:
Until now, we have used gcc/g++ for compilation. However, more recently we have consistently been running into a bug in gcc's version of libsanitizer in which it had a race condition between alignment checks and allocation patterns for dynamic thread local storage memory, caused by dlopen calls. Because we support gobject multi-language bindings and our integration test harness is thus in python, we experience quite a lot of dlopen calls. Here are just a few issues filed for this bug:
https://github.com/google/sanitizers/issues/1322
https://github.com/google/sanitizers/issues/1409
https://github.com/google/sanitizers/issues/1170

This bug was fixed for LLVM's compiler-rt 3 years ago here: https://reviews.llvm.org/D147459

however, gcc's libsanitizer (which is a downstream fork of compiler-rt) has yet to pick this fix up. Whether it ever will remains to be seen. For now, we will use llvm/clang 18 by default.

This commit additonally removes gcc/g++ 11 from the Dockerfile as it was originally added for Matter but v1.5.1.0 supports the system default of gcc/g++ 13 so this is no longer needed.


docker/version
```diff
@@ -1 +1 @@
-2.9
+llvm
```
tleacmcsa: dont forget to change this...

tleacmcsa (dismissed):

docker/version
cleithner-comcast: Thank you for reminding me!

tleacmcsa: Also note that #201 is using 3.0 currently.  either this or that can change, i dont care.  just FYI.
cc @Lakshmi97-velampati

testing/py_test.sh
```diff
@@ -35,19 +35,65 @@
...
 fi
 
-LD_PRELOAD=$(gcc -print-file-name=libasan.so) pytest "$@"
+case "$TOOLCHAIN" in
+    clang)
+        ASAN_LIB=$(clang -print-file-name=libclang_rt.asan-x86_64.so)
+        ;;
+    gcc)
+        ASAN_LIB=$(gcc -print-file-name=libasan.so)
+        ;;
```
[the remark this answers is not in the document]
cleithner-comcast: I'm not worried about that, thanks though.

tleacmcsa (dismissed):

docker/version
cleithner-comcast: I think (hope) this is going to make it in first as other PRs are running into the flaky race condition so I have set the version to 2.10.

docker/patches/otbr/0005-fix-clang-vla-warning.patch
```diff
@@ -0,0 +1,17 @@
...
+index 1234567..abcdefg 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -34,6 +34,10 @@ project(openthread-br VERSION 0.3.0)
+ 
+ add_compile_options(-Wall -Wextra -Werror -Wfatal-errors -Wuninitialized -Wno-missing-braces)
+ 
++if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
++    add_compile_options(-Wno-vla-cxx-extension)
++endif()
```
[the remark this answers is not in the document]
cleithner-comcast: It doesn't.

rchowdcmcsa (approved):

kfundecmcsa (approved):

PR #228, opened 2026-06-09
feat(sbmd): explore JS-based SBMD driver format

tleacmcsa:
The current YAML-based SBMD format has exposed shortcomings when tackling advanced features like door lock event handling and camera WebRTC protocol drivers. The core gap: SBMD drivers are glue between Barton resource operations and asynchronous device-side messaging, but the path for incoming device data (attribute changes, events, command responses) was not well considered in the YAML approach.

This commit introduces an exploratory JS-based SBMD format:

- Replace YAML-with-embedded-JS with a single pure JavaScript file, enabling better tooling, authoring, and test fixture support.
- Declare all driver metadata, resources, endpoints, and device-side handlers in one SbmdDriver() registration object.
- Cleanly separate Barton resource handlers (read/write/execute) from Matter device-side handlers (attribute, event, command).
- Support flexible handler dispatch: single, multi, and wildcard matching for attributes, events, and commands.
- Introduce a builder pattern (SbmdUtils.result()) for composing atomic multi-operation responses from handler functions.
- Add runtime-managed constants, optional handler inputs for pre-fetching Barton resources and Matter attributes, and persistent/transient key/value storage.


core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+  var isLocked = (value === 1);
+
+  return SbmdUtils.result()
+    .updateResource(isLocked ? "true" : "false");
+}
+
+/** Read the current identify string. */
+function readIdentify(args) {
+  return SbmdUtils.result()
+    .read(IDENTIFY_CLUSTER, 0x0000);
```
cleithner-comcast: Does this mean read from the device instead of from the data cache? I'm confused how this is intended to be used.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+  var tlvBase64 = SbmdUtils.Tlv.encodeStruct({ IdentifyTime: parseInt(args.value) }, schema);
+
+  return SbmdUtils.result()
+    .write(IDENTIFY_CLUSTER, 0x0000, tlvBase64);
+}
+
+/** Reboot the device via a General Diagnostics cluster command. */
+function executeReboot(args) {
+  return SbmdUtils.result()
+    .invoke(GENERAL_DIAGNOSTICS_CLUSTER, 0x0000, null, {});
```
cleithner-comcast: How should the native side interpret no callback? I am on the fence about requiring the callback parameter. Should drivers be held accountable for translating matter operation errors for error responses back into the native layer?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+// Attribute handler implementations
+// ===========================================================================
+
+/** Handle LockState attribute — same transform as the resource read. */
+function handleLockStateAttribute(args) {
+  var value = SbmdUtils.Tlv.decode(args.tlvBase64);
+  var isLocked = (value === 1);
+
+  return SbmdUtils.result()
+    .updateEndpointResource(LOCK_ENDPOINT, "locked", isLocked ? "true" : "false");
```
cleithner-comcast: Can we just do updateResource and have endpoint id be optional? Also, don't forget to support metadata with the resource model operations.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+  var value = SbmdUtils.Tlv.decode(args.tlvBase64);
+  var currentLocked = args.resources[LOCK_ENDPOINT + "/locked"];
+
+  if (args.attributeId === 0x0002) {
+    return SbmdUtils.result()
+      .updateEndpointResource(LOCK_ENDPOINT, "actuatorEnabled", value ? "true" : "false");
+  } else if (args.attributeId === 0x0003) {
+    return SbmdUtils.result()
+      .updateEndpointResource(LOCK_ENDPOINT, "doorState", String(value))
+      .log("doorState changed while locked=" + currentLocked);
```
cleithner-comcast: Logging is interesting. I feel like logging as a function result might not be useful. But providing a log hook would allow calling back into the native side. Hmm

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+
+  if (!actuatorEnabled) {
+    return SbmdUtils.result()
+      .log("lock operation ignored — actuator disabled");
+  }
+
+  if (opType === 0) {
+    return SbmdUtils.result()
+      .updateEndpointResource(LOCK_ENDPOINT, "locked", "true")
+      .setPersistentKey("lastLockOperation", "lock");
```
cleithner-comcast: I would suggest replacing Key with Data in these routines. Saving off keys don't make no sense.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+        },
+        lock: {
+          type: "function",
+          execute: executeLockAction,
+        },
+        unlock: {
+          type: "function",
+          execute: executeLockAction,
+        },
+      },
```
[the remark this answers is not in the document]
tleacmcsa: @copilot add "read" mode to these also

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: oops, thats bogus. fixed.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: yeah, lets noodle on this.  errors could be generic (invalid argument, not authorized, etc.) or they could be command specific (user table full).

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -210,34 +222,34 @@ function readLockedState(args) {
...
   return SbmdUtils.result()
-    .read(IDENTIFY_CLUSTER, 0x0000);
+    .read(IDENTIFY_CLUSTER, ATTR_IDENTIFY_TIME);
 }
 
 /** Write a new identify duration. */
 function writeIdentify(args) {
   var schema = { IdentifyTime: { tag: 0, type: "uint16" } };
-  var tlvBase64 = SbmdUtils.Tlv.encodeStruct({ IdentifyTime: parseInt(args.value) }, schema);
+  var tlvBase64 = SbmdUtils.Tlv.encodeStruct({ IdentifyTime: parseInt(args.resource.input) }, schema);
```
cleithner-comcast: Is args.resource.input the write resource value? I'm quickly realizing we need strongly types function arguments so authors can see what these properties are.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -161,22 +166,20 @@ SbmdDriver({
   },
 
   commandHandlers: {
```
cleithner-comcast: What direction are these commands? I would expect a driver to have handles for server cluster commands we host, but why would it need command handlers for reaching out to the device? I would expect that would need to go through a formal driver interface like an executeResource handler.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: I dont think we can leave endpoint id out.  There is no relationship between this attribute and a barton resource here.

Added an example metadata payload here.  Good catch.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: done.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: yeah, that would be input to a write or execute resource operation.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: These are the response handlers.  see their names (handleGetCredentialStatusResponse) and command ids (CMD_GET_CREDENTIAL_STATUS_RESP) and leading comment (// Single command response with supplementary attribute context.).

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: Yeah, I guess I mean if we allow the callback to be optional, we now need to add default processing in the native layer so things like resource executes propogate errors or OKs to client calls. This might be annoying to try and guess, so there may be some benefits to requiring a callback from the driver so the driver has to be explicit about the response.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: I didn't say leave out endpoint id. I said let's not have two functions for device resources and endpoint resources. Let's have one function for updating resources and have endpoint id be an optional parameter to differentiate between device and endpoint resources.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: How is that supposed to work though? I execute a resource, the resource execute handler in the driver needs to run a command, and it's just implied that that command response will go to these functions? I think when the driver returns an invoke for a command, the command result should go to the callback specified, not to declared command handlers.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -191,6 +215,15 @@ SbmdDriver({
```
cleithner-comcast: This is hard to understand because there's no defined contract for what the attributes looks like. I wonder if we should provide some strong types in SbmdUtils for non-custom supplement args.

Also, are constants defined in the original json object usable in functions? Maybe they need to be provided via args.constants?

I forgot to post this on my last set of comments.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,412 @@
...
+  schemaVersion: "4.0",
+  driverVersion: "1.0",
+  name: "Door Lock",
+
+  constants: {
+    LOCK_ENDPOINT: "1",
+    DOOR_LOCK_CLUSTER: 0x0101,
+    IDENTIFY_CLUSTER: 0x0003,
+    GENERAL_DIAGNOSTICS_CLUSTER: 0x0033,
+  },
```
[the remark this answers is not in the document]
tleacmcsa: We are actively preventing dynamic storage like var for now.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: possibly.  i havent thought this through yet.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: im not following your comment about the attributes contract.   please elaborate.

Yes, I need to add the constants to args.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+    },
+
+    // Multiple attributes — several attributes share one handler.
+    // Pre-fetches the current "locked" resource value as supplementary context.
+    lockActuator: {
+      clusterId: DOOR_LOCK_CLUSTER,
+      attributeIds: [ATTR_ACTUATOR_ENABLED, ATTR_DOOR_STATE],
+      supplements: {
+        resources: [LOCK_ENDPOINT + "/locked"],
+      },
```
rchowdcmcsa: The comment says this is pre-fetched. Is that a synchronous read from the cached resources? What if the supplement isn't cached yet before the handler is invoked? Can that happen?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+  //     handler runs.
+  // =========================================================================
+
+  attributeHandlers: {
+    // Single attribute — one specific attribute, one handler
+    lockState: {
+      clusterId: DOOR_LOCK_CLUSTER,
+      attributeId: ATTR_LOCK_STATE,
+      handler: handleLockStateAttribute,
+    },
```
rchowdcmcsa: So before, we could have a read mapper that updated resource values using attribute reports, or we could have an event mapper that did it by reading the event payload, and the event mapper also had the `seedFrom` field so that the resource value could be populated on startup by reading the attribute (after which point we just relied solely on the event).

How would it work now? We have this attribute handler for the lock state, and the handler updates the locked resource. We also have an event handler down below (lockOperation) that also updates the locked resource. Which one do we listen to for resource updates? Both? We'd be doing a double update every time then, no? (except under poor network conditions where attribute reports might not make it through I suppose.)

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+      handler: handleActuatorAttributes,
+    },
+
+    // Wildcard — catch-all for any attribute on a cluster
+    lockDiagnostics: {
+      clusterId: DOOR_LOCK_CLUSTER,
+      attributeId: "*",
+      handler: handleLockDiagnostics,
+    },
+  },
```
rchowdcmcsa: So when the lock state attribute changes, do both the lockDiagnostics and the lockState attribute handlers fire? In what order? Is the onus just on the author to ensure that the functionality of any wildcard attribute or event handlers do not step on the toes of the non-wildcard handlers that are implemented?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+
+  if (!actuatorEnabled) {
+    return SbmdUtils.result()
+      .log("lock operation ignored — actuator disabled");
+  }
+
+  if (opType === 0) {
+    return SbmdUtils.result()
+      .updateEndpointResource(args.constants.LOCK_ENDPOINT, "locked", "true")
+      .setPersistentData("lastLockOperation", "lock");
```
rchowdcmcsa: On chaining.
Can you chain two .updateEndpointResource() calls for different resources? I assume yes.
What if you chain two updates to the same resource? I assume last-one-wins applies since the builder goes in sequential order. Do we just trust authors not to do this, or should we explicitly make this 'not allowed'?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+  } else if (opType === 1) {
+    return SbmdUtils.result()
+      .updateEndpointResource(args.constants.LOCK_ENDPOINT, "locked", "false")
```
rchowdcmcsa: We're no longer relying on an alias (or similar) that is declared somewhere in the file and validated by the parser. We're using bare strings. What happens if I accidentally write "lokced"? Silent failure? Runtime error?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,447 @@
...
+/**
+ * Handle DoorLockAlarm + LockUserChange events.
+ *
+ * Stores the alarm code transiently (auto-cleaned after 300s) for
+ * short-lived diagnostics, and persists the total alarm count across
+ * reboots.
+ *
+ * getPersistentData(name) retrieves a value from non-volatile storage
+ * (returns null if not set).  setTransientData(name, value, ttlSecs)
+ * stores in memory with automatic cleanup after ttlSecs seconds.
```
rchowdcmcsa: Would the non-volatile storage mentioned here be updated immediately whenever setPersistentData() is called on the builder? I guess I'm just wondering if the possibility of a handler reading stale data is something we have to worry about.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: Good thinking.  All of the handler functions take the same argument shape.  So it is possible for the lockOperation event handler, the  lockState attribute report handler, and the resource read handler to all point to the same function -- they could return an operation to update the locked resource value.  The catch is that we only wanted to use the attribute for the initial value at east startup, then use events after that.  We can do that by setting a "flag" (SbmdUtils.setTransientData) that indicates we got the initial value from the attribute already and then ignore them from there.  I'm not suggesting we do it exactly like that, but that its an option.  We should come up with the right solution here, then update the example.  What do you think?

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: the cache is populated at pairing and startup.  It is guaranteed that any readable attribute is cached locally before this gets invoked.

tleacmcsa: any -> all.  We do a wildcard subscribe on all attributes on all clusters on all endpoints of every device.  The spec says that will return all readable attributes immediately and then upon change.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: right now that sequencing is not specified.  If we need to make it a contract, we can.  Right now they would all fire in any order.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: When I define the device model at the top, I want to optionally specify how a resource gets its startup value. That itself could be a handler function that is separate from the read/write/execute function handlers (regular operations).

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: the output from builder is a sequence of operations.  they will be evaluated in order.  If the sequence starts with a long running operation like "fetch users from door lock" and then does "setPersistentData", that data wont be set until some time after the fetch users from door lock operation completes, but before the next operation in the sequence.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: IMO updating a resource is NOT the same as a write resource result (I think these should be separate APIs in the result object). So in a handleWriteLockState function, it might update locked or last interact time or something else all in the same function but in order for the native side to know the final result, it needs either an error result or a write resource result. In this model, updateResource results are completely arbitrary, can be 0....N in one return, and are executed in order sepcified.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
cleithner-comcast: In this particular example, it's a resource value which is an arbitrary string. If a dev messes it up, it should be caught by tests. Aliases wouldn't be relevant here. However, Thomas does introduce constants into this PR so I would think string constants could be defined as well. Whether or not an alias is needed would only be determined by the json initializer at the top. I can't say if we still need them or not at this time.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
rchowdcmcsa: @tleacmcsa I think I prefer Christian's idea of there being a handler function that is separate from read/write/event/etc., like an initialize value handler or something of the sort. This would sort of mirror how the old model does it where seedFrom is a specific mapper that is designed solely to be invoked once for the initial resource population. I feel like this is less complicated than checking a flag (not sure how it'd look when implemented though).

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: Yeah, I think a string constant would help this scenario.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: I am missing the reason why updating a resource as part of an operation output would be different than the final result/operation on a writeResource.  Ultimately they both need an operation to set a resource value and i dont see why they wouldnt be the same.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: yeah, i like it.  will update.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: Meh, I liked the explicit two version approach but I did make the change you requested.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: fixed.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
rchowdcmcsa: I think that's ok for now, just something we need to be aware of when authoring these drivers.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,458 @@
...
+ * version 3
+ */
+
+SbmdDriver({
+  schemaVersion: "4.0",
+  driverVersion: "1.0",
+  name: "Door Lock",
+
+  constants: {
+    LOCK_ENDPOINT: "1",
```
kfundecmcsa: Nit: `ATTR_*`, `EVT_*`, `CMD_*`, and `RES_*` all use a category prefix, but endpoints and clusters use a suffix (`LOCK_ENDPOINT`, `DOOR_LOCK_CLUSTER`). Might be worth making this consistent, e.g. `EP_LOCK`, `CL_DOOR_LOCK`, so all constant categories are prefix-namespaced and easy to filter/autocomplete by type.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
```diff
@@ -0,0 +1,458 @@
...
+// Resource handler implementations
+//
+// All handlers receive the same args shape:
+//   args.attribute  — { clusterId, attributeId, value }  (attribute handlers)
+//   args.event      — { clusterId, eventId, data }       (event handlers)
+//   args.command    — { clusterId, commandId, data }     (command handlers)
+//   args.resource   — { resourceId, input }              (resource handlers)
+//   args.supplements.attributes — { [clusterId]: { [attributeId]: value } }
+//   args.supplements.resources  — { "endpointId/resourceId": value }
+// Result builder methods:
```
kfundecmcsa: As this list grows (timers? error reporting?), a flat builder could become hard to discover/navigate without types. Might be worth considering namespacing the methods by concern, e.g. `.matter.invoke()`, `.barton.updateResource()`, `.storage.persist()`, or at minimum establishing a naming convention that groups related methods.

kfundecmcsa (commented):
just a couple minor suggestions after reviewing all the conversations and slack chat

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: good idea.  done.

core/deviceDrivers/matter/sbmd/specs/door-lock.sbmd.js
tleacmcsa: good idea.  done.

docs/SBMD.md
```diff
@@ -1,1349 +1,1608 @@
...
+SBMD eliminates the need to write per-device-type native C/C++ drivers. New Matter
+device types can be supported by adding a specification file — no firmware rebuild
+or redeployment required.
+
+### 1.1 Goals
+
+- **Single-file drivers**: One `.sbmd.js` file fully defines a device driver —
+  metadata, resource declarations, device-side handler registrations, and all
+  handler implementations.
+- **No `var` in driver scope**: Driver authors never allocate memory directly.
```
kfundecmcsa: I like this constraint, makes total sense. Might be worth adding a brief "why" here so future driver authors understand the motivation and don't try to work around it.

docs/SBMD.md
```diff
@@ -1,1349 +1,1608 @@
...
+2. **Registration**: Each `SbmdDriver()` call registers a driver with its metadata,
+   resource declarations, and handler functions.
+3. **Device claiming**: When a Matter device is commissioned, a two-pass process
+   selects the driver: vendor-specific drivers (matched by `vendorId`/`productId`)
+   are tried first, then generic device-type drivers.
+4. **Resource binding**: The driver creates Barton resources based on the endpoint
+   and resource declarations, gated by alias prerequisites.
+5. **Runtime operations**:
+   - **Attribute report** → attribute handler → result builder → resource update
+   - **Resource read** → read handler (with supplements) → result builder → value
```
kfundecmcsa: "supplements" is used here for the first time but isn't defined until 4.12, might be worth linking to that section or adding a brief explanation like "(pre-fetched device data; see 4.12)".

docs/SBMD-v3-legacy.md
```diff
@@ -0,0 +1,1349 @@
...
+Specification-Based Matter Drivers (SBMD) is a device driver framework that enables
+Barton to support Matter devices through declarative YAML specification files rather
+than compiled C/C++ code. This approach facilitates:
+
+- **Rapid device type support**: Add new Matter device types without code changes
+- **Dynamic extensibility**: Deploy new device support without firmware updates
+- **Simplified maintenance**: Declarative specifications are easier to review and maintain
+- **Reduced complexity**: Eliminate per-device-type native code compilation
+
+### 1.2 Historical Context
```
kfundecmcsa, line 39: I like this block from the OG SMBD, particularly calling out Barton's resource-based model and its relationship to device-specific interfaces, and a bit of the history - but looks like it's been removed in the newer version. May consider plopping this back in place. I found it interesting and useful.

kfundecmcsa (commented):
Great stuff! Made it part way through, I'll finish the rest soon

docs/SBMD.md
tleacmcsa: added a blurb.

docs/SBMD.md
tleacmcsa: added link to section.

docs/SBMD-v3-legacy.md
tleacmcsa: added back.

docs/SBMD.md
```diff
@@ -1,1349 +1,1627 @@
...
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `clusterId` | number | yes | Matter cluster ID. |
+| `attributeId` | number | conditional | Attribute ID. Mutually exclusive with `eventId` and `commandId`. |
+| `eventId` | number | conditional | Event ID. Mutually exclusive with `attributeId` and `commandId`. |
+| `commandId` | number | conditional | Command ID. Mutually exclusive with `attributeId` and `eventId`. |
+| `type` | string | no | Matter data type (documentation only, ignored by runtime). |
+
+Aliases serve three purposes:
```
cleithner-comcast, line 262: are aliases still relevant? Are prerequisites still relevant? We didn't walk through those in the examples.

docs/SBMD.md
```diff
@@ -1,1349 +1,1627 @@
...
-| Alias type | Check performed |
-|------------|----------------|
-| `attribute` alias | Cluster **and** attribute must be present in the device's data cache |
-| `event` alias | Cluster must be present in the device's data cache |
+**Seed vs Read**
 
-All listed prerequisites must be satisfied for the resource to be registered. If
-any prerequisite fails and the resource is required (no `optional: true`), the
-driver aborts commissioning. If the resource is optional, it is silently skipped.
+- `seed` runs when the device is first discovered **and** each time Barton
```
cleithner-comcast, line 496: is seed still the terminology? We jumped from examples to here before showing an example with initial resource value handler support.

docs/SBMD.md
```diff
@@ -1,1349 +1,1627 @@
...
+### 4.9 Attribute Handlers
+
+Attribute handlers process incoming Matter attribute reports from the device.
+
+```js
+attributeHandlers: {
+  // Alias form — resolved to cluster + attribute from the aliases section
+  handlerName: {
+    aliases:      string[],          // alias names (mutually exclusive with clusterId)
+    supplements:  { ... },           // optional: pre-fetched data
```
cleithner-comcast, line 531: Are these copies? Are these immutables?

docs/SBMD.md
tleacmcsa: Yes, they both have their places.  They were missed in the examples, but found as regressions in the new stuff.

docs/SBMD.md
tleacmcsa: I think you missed the examples at the end of the doc.  Yes, seed is still the terminology and there is an example on line 1356

docs/SBMD.md
```diff
@@ -1,1349 +1,1627 @@
...
+          type: "boolean",
+          modes: ["read"],
+          seed: {
+            supplements: {
+              attributes: ["lockState"],
+            },
+            handler: seedLockedResource,
+          },
+        },
+        [RES_LOCK]: {
```
cleithner-comcast, line 1693: Not super fan of using aliases for resource keys here. I don't know the difference between RES_LOCK and RES_LOCKED without going back and forth with the constants.

docs/SBMD.md
tleacmcsa: immutable copies.  will clarify.

tleacmcsa: done

docs/SBMD.md
cleithner-comcast: Yes i did

docs/SBMD.md
```diff
@@ -1,1349 +1,1629 @@
...
+  var actuatorEnabled = args.supplements.attributes.actuatorEnabled;
+
+  if (!actuatorEnabled) {
+    return SbmdUtils.result()
+      .log("lock operation ignored — actuator disabled");
+  }
+
+  if (opType === 0) {
+    return SbmdUtils.result()
+      .barton.updateResource(EP_LOCK, RES_LOCKED, "true")
```
cleithner-comcast: Are we wanting ALL barton operations to go here? If we add support for using barton property system, would the operation go under "barton" namespace alongside updateResource which is a device specific operation? I wonder if updateResource should be under the namespace "dataModel".

docs/SBMD.md
```diff
@@ -1,1349 +1,1629 @@
...
-4. Validate YAML syntax before deployment
-5. Test scripts with unit tests before integration
+  return SbmdUtils.result()
+    .device.write(CL_IDENTIFY, ATTR_IDENTIFY_TIME, tlvBase64);
+}
 
-## 9. File Deployment
+function executeReboot(args) {
+  return SbmdUtils.result()
+    .device.invoke(CL_GENERAL_DIAGNOSTICS, CMD_REBOOT, null, {});
```
cleithner-comcast: what does a null paylod mean for invoke?

docs/SBMD.md
```diff
@@ -1,1349 +1,1629 @@
...
-```
+| Parameter | Type | Description |
+|---|---|---|
+| `endpoint` | string | Endpoint ID (use an `EP_*` constant). |
+| `resource` | string | Resource name (use a `RES_*` constant). |
+| `value` | string | New resource value. |
+| `metadata` | string | Optional. JSON string of metadata to attach to the update. |
+
+The runtime distinguishes the two-arg vs three-arg form by the first argument:
+endpoint IDs are numeric strings, resource names are not.
```
rchowdcmcsa: Is it worth documenting this as an explicit constraint somewhere around section 4.2? e.g. "Endpoint ID constants MUST be numeric strings; resource name constants MUST NOT be."

docs/SBMD.md
```diff
@@ -1,1349 +1,1629 @@
...
+- No global `var` declarations are permitted at file scope. The runtime may
+  reject files that declare `var` outside of function bodies.
+- `SbmdUtils` and `SbmdDriver` are the only runtime-provided globals (aside
+  from injected constants and standard JavaScript built-ins).
+
+### 9.4 Handler Dispatch Order
+
+When an incoming attribute/event/command matches multiple registered handlers:
+
+1. **Specific handlers** (single `attributeId`/`eventId`/`commandId`) fire first.
```
rchowdcmcsa, line 1261: If there are two specific handlers that match one event, is it random? Or will we enforce a constraint that says there can only be one specific handler per event, etc.?

docs/SBMD.md
```diff
@@ -1,1349 +1,1629 @@
...
+        [RES_UNLOCK]: {
+          type: "function",
+          execute: executeLockAction,
+        },
+        [RES_ACTUATOR_ENABLED]: {
+          type: "boolean",
+          prerequisites: ["actuatorEnabled"],
+          optional: true,
+          modes: ["read"],
+        },
```
rchowdcmcsa, line 1707: We have some resources with neither a read nor seed handler, like this one. They just rely on an attribute handler to eventually populate the resource. Which is fine, but I wonder if it's worth documenting why we sometimes want the ability to fetch a value and why in other cases we're satisfied just reacting to attribute reports as they come in.

docs/SBMD.md
tleacmcsa: Their use is optional here.  I'll add some more basic examples.

tleacmcsa: previously it was "locked" and "lock" strings in place of those.  Dunno why that wasnt just as confusing...?

docs/SBMD.md
tleacmcsa: dataModel is a bit loose.  How about "resource"?  SbmdUtils.result().resource.update(...)

docs/SBMD.md
tleacmcsa: the matter command reboot on the general diagnostics cluster is sent without a payload because it does not take one.

docs/SBMD.md
tleacmcsa: endpoint IDs for barton are strings, not necessarily numeric.  We have been using numbers, but that is not required per architecture.

docs/SBMD.md
tleacmcsa: I think we touched on this yesterday.  If you set it up so that two handlers will match, there is no guarantee on the order of execution.  It sounds like a bad idea.

docs/SBMD.md
tleacmcsa: added a section "No handler (event-driven resources)" that talks about this.

docs/SBMD.md
```diff
@@ -906,61 +964,109 @@ These are also available as standalone read accessors:
...
+    .success();
 }
 ```
 
+Every result chain must end with an explicit terminal. A chain with no terminal
+is a **runtime error**.
+
 ### 7.6 Error
 
 #### `error(message)`
```
cleithner-comcast, line 1045: I think CHIP Error codes should be passed to the driver. I don't wanna assume drivers cannot conditionally react.

docs/SBMD.md
```diff
@@ -906,61 +964,109 @@ These are also available as standalone read accessors:
...
+      .log("rejected invalid write: " + value)
+      .error("invalid value: " + value);
   }
 
   var commandId = (value === "true") ? CMD_ON : CMD_OFF;
 
   return SbmdUtils.result()
-    .device.invoke(CL_ON_OFF, commandId, null, {});
+    .device.sendCommand(CL_ON_OFF, commandId, null, {})
+    .success();
```
cleithner-comcast: a no arg success does not seem like a ubiquitous terminator. Don't read resource handlers need to return a value in the result? Don't write resource and execute resource have optional data they can provide back to the client? Don't write resource handlers need to be able to provide metadata for the event update of the new resource value?

docs/SBMD.md
cleithner-comcast: That is equally confusing, but the main hangup for me is when I look at a data model dump for a device and then at the model in the driver, I can't just match up stirng literals, I have to use a mapper. It's not a big deal, and I may find it's never actually an issue.

docs/SBMD.md
cleithner-comcast: Well presumably a driver would want to update metadata on a device as well.

docs/SBMD.md
cleithner-comcast: Ah, okay I understand now.

docs/SBMD.md
tleacmcsa: good idea. added.

docs/SBMD.md
tleacmcsa: all handler functions just run code now, they arent limited to the previous 'translator only' mechanism which expected them to return data that goes into the resource.  if a read handler should update an associated resource, it needs to call updateResource to do so, then indicate success() if it is done, or chain on another operation if desired.

docs/SBMD.md
rchowdcmcsa: Doesn't that make "The runtime distinguishes [...] by the first argument: endpoint IDs are numeric strings [...]" a fragile convention to document here?

docs/SBMD.md
cleithner-comcast: This doesn't really cover my question. I understand philosophically that's the case but I think there are gaps in these use cases that I noted. I would think that in the same way that your handler will get different args depending on what it is associated with, what's needed from the handler in the result also varies.

For example, in the camera data model lets say a client executes the createSession resource. The handler for this would want to return this session id to the client. Here's what it looks like in v3:

```resources:
      # Creates a new session, returns sessionId
      - id: "createSession"
        type: "function"
        prerequisites: none
        mapper:
          execute:
            script: |
              var sessions = SbmdUtils.SessionManager.getForDevice(sbmdCommandArgs.deviceUuid);
              var sessionId = sessions.create({});
              return SbmdUtils.Response.value(sessionId.toString());
```

How do we do this in v4?

tleacmcsa:
```
...
        "createSession": {
          type: "function",
          execute: createSessionAction,
        },

...

function createSessionAction(args) {
  return SbmdUtils.result()
    .session.create(
      handler: function(args) {
        return SbmdUtils.result()
          .barton.updateResource("createSession", args.session.id)
          .success();
      },
      onError: function(args) {
        return SbmdUtils.result()
          .log("create session failed: " + args.error.message)
          .error(args.error.message);
      }
    );
}
```

cleithner-comcast: I would not expect session code to live in SbmdUtils anymore in v4, but I guess that's besides the point. Does updateResource do something different for execute resources than ones that retain a value?

docs/SBMD.md
tleacmcsa: indeed! fixed.

docs/SBMD.md
tleacmcsa: good point. done.  added setMetadata too

docs/SBMD.md
tleacmcsa: Thats fine, i wanted to show it here since it would need to be async now.

I'm not sure I follow your question about updateResource.  It is just an operation and has no direct relationship with a resource (read, execute, whatever).  A handler that gets called when a resource is read is very likely going to call updateResource (not required though).  Remember, the handlers are not returning values for a particular request anymore.  They return operations.

Similarly, a handler that is called when a resource is executed probably wont call updateResource, but it could I suppose.

docs/SBMD.md
```diff
@@ -1,1349 +1,1939 @@
...
-Using aliases eliminates duplication — a cluster/attribute pair is defined once and
-referenced by name wherever it is needed.
+See [4.8.1 Resource Declaration](#481-resource-declaration) for the full schema.
 
-### 3.5 Reporting Configuration
+### 4.8 Endpoints
 
-A single wildcarded attribute reporting configuration is maintained on the device.
-These settings allow configuration on the min/max intervals.
+Each endpoint carries a profile and its own set of resources.
```
kfundecmcsa, line 388: The v3 doc had an explicit callout (section 3.6) that endpoints are Barton data model concepts, not Matter endpoints. That distinction seems to have gotten lost in the v4 rewrite, might want to add that back in here.

tleacmcsa: Good catch!  Fixed.

docs/SBMD.md
cleithner-comcast: A client, when they call execute resource, can optionally receive data back that ISN'T the value of the resource (because executable resources have no value). This could be command result data, it could be custom data from the driver, whatever. This is a feature of the device data model. I don't see that feature supported in sbmd v4. I don't see a way to tie a synchronous client execute resource operation with a value from the driver. If you assert that updateResource, for some reason, can now handle this case, it opens up the problem of any random handler being able to use updateResource on an executable resource and the behvaior there being undefined.

tleacmcsa: ahh!  I forgot about that capability.  Thanks!  I'll look into it

tleacmcsa: take a look at the commit i just added.  it has details in its message.

cleithner-comcast: Okay I'm satisfied with this support. I do think we have a gap with event metadata (ex: resource update event metadata). Not sure if we want to tackle that right now.

tleacmcsa: I added optional metadata to .success(value, metadata?) ... pushed.

docs/SBMD.md
```diff
@@ -1065,7 +1070,7 @@ ultimately resolve to success or failure. The rules are:
 | Chain contains | Outcome |
 |---|---|
 | `.success()` | Success. All preceding operations in the chain execute. |
-| `.success(value)` | Success. Sets the result value for the resource operation (updates the resource for read/seed/write; returns to caller for execute). Only valid on resource handlers and deferred response handlers. |
+| `.success(value)` | Success. Sets the result value for the resource operation (updates the resource for read/seed/write; returns to caller for execute). Accepts optional second `metadata` argument (JSON string) for resource updates. Only valid on resource handlers and deferred response handlers. |
```
cleithner-comcast: Are the "Chain contains" column entries supposed to have all arguments in the signatures? For example, should it be .success(value, metadata)? Some of the others also list as terminal, but .success/.error are not marked as terminal. Maybe some gaps in documentation in these blocks, but earlier documentation (example: section 7.5) is sufficient.

cleithner-comcast (dismissed):

docs/SBMD.md
tleacmcsa: ahh.   take a look now.

docs/SBMD.md
```diff
@@ -1,1349 +1,1969 @@
...
-- **Barton side**: Resource values are represented as **strings**. All Barton resource
-  reads return strings, writes accept strings, and function arguments/responses are strings.
+| Field | Type | Description |
+|---|---|---|
+| `args.supplements.attributes` | `{ [aliasName]: value }` | Pre-fetched attribute values, keyed by alias name. |
+| `args.supplements.resources` | `{ [path]: value }` | Pre-fetched resource values. Keys are `"endpointId/resourceName"` or `"resourceName"`. |
 
-- **Matter side**: Data is encoded as **TLV** (Tag-Length-Value) binary format for
-  over-the-air communication with devices.
+#### Deferred handler context (present on response/error handlers)
```
kfundecmcsa, line 700: I was confused by this section for a while - "Deferred handler context" introduces a concept that isn't explained until sections 6–7. A one-liner before the table would help, something like: "When a handler defers completion via `.device.requestCommand()` (see Section 6), the response/error callback receives these additional fields:" or something like that

tleacmcsa: good idea.  done.

cleithner-comcast (dismissed):

docs/SBMD.md
```diff
@@ -1,1349 +1,1979 @@
...
+| `.device.readAttribute()` | no | Defers to response `handler` or `onError`, which must provide a terminal. See [7.2](#72-device-interaction--device). |
+| *(none)* | — | **Runtime error.** Every chain must end with an explicit terminal. |
+
+**Single path to terminal**: A result chain must contain exactly **one** path to
+a terminal. A chain must not include multiple deferred operations
+(`requestCommand`, `readAttribute`) because each defers to its own handler,
+creating ambiguous completion. The runtime rejects chains with more than one
+deferred operation.
+
+**Timeout resolution**: Per-operation `timeoutMs` > `matter.defaultTimeoutMs` > system default.
```
kfundecmcsa: The `>` notation here reads a bit ambiguously (math comparison vs. precedence). Maybe: "Timeout priority (highest to lowest): per-operation `timeoutMs` > `matter.defaultTimeoutMs` > system default."

tleacmcsa:
this PR was not intended to be merged, only for review.  review done, implementation on the feature starting on new branch (tlea/dev/sbmd-v4)

PR #235, opened 2026-06-22
feat(observability): add metrics abstraction with in-mem and noop backends

tleacmcsa:
Add a lightweight observability framework with pluggable backends:
- inmemory: tracks counters and timing histograms for SBMD operations
- noop: zero-cost stub when metrics are disabled

Includes CMake option BCORE_OBSERVABILITY_BACKEND to select the backend at build time, header-only metric macros, and a CMocka unit test.


config/cmake/options.cmake
```diff
@@ -176,6 +176,10 @@ bcore_option(NAME BCORE_BUILD_THIRD_PARTY_BARTON_COMMON
            DESCRIPTION "Build the third-party BartonCommon component"
            ENABLE)
 
+set(BCORE_OBSERVABILITY_BACKEND "inmemory" CACHE STRING "Observability backend (none, inmemory)")
```
cleithner-comcast: nit: I would just call it "memory"

config/cmake/options.cmake
```diff
@@ -176,6 +176,10 @@ bcore_option(NAME BCORE_BUILD_THIRD_PARTY_BARTON_COMMON
...
+set_property(CACHE BCORE_OBSERVABILITY_BACKEND PROPERTY STRINGS none inmemory)
```
cleithner-comcast: Why not use bcore_string_option?

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+} GaugeDataPoint;
+
+/* ------------------------------------------------------------------ */
+/* Histogram buckets                                                  */
+/* ------------------------------------------------------------------ */
+
+/* Default bucket boundaries matching OpenTelemetry SDK defaults */
+static const double kHistogramBounds[] = {
+    0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000};
+static const int kNumBounds = sizeof(kHistogramBounds) / sizeof(kHistogramBounds[0]);
```
cleithner-comcast: Should just use ARRAY_SIZE

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+
+typedef struct HistogramDataPoint
+{
+    AttrSet attrs;
+    uint64_t count;
+    double sum;
+    double min;
+    double max;
+    uint64_t buckets[sizeof(kHistogramBounds) / sizeof(kHistogramBounds[0]) + 1];
```
cleithner-comcast: should use kNumBounds

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    AttrSet attrs;
+    int64_t value;
+} GaugeDataPoint;
+
+/* ------------------------------------------------------------------ */
+/* Histogram buckets                                                  */
+/* ------------------------------------------------------------------ */
+
+/* Default bucket boundaries matching OpenTelemetry SDK defaults */
+static const double kHistogramBounds[] = {
```
cleithner-comcast: k prefix bleeding

core/src/observability/memory/observabilityMemory.c
```diff
@@ -0,0 +1,835 @@
...
+
+typedef struct InstrumentBase
+{
+    InstrumentType type;
+    char *name;
+    char *description;
+    char *unit;
+    int refCount;
+    struct InstrumentBase *next; /* linked list in global registry */
+} InstrumentBase;
```
cleithner-comcast, line 208: I would like to see GSList (or GList) used.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    INSTRUMENT_HISTOGRAM
+} InstrumentType;
+
+typedef struct InstrumentBase
+{
+    InstrumentType type;
+    char *name;
+    char *description;
+    char *unit;
+    int refCount;
```
cleithner-comcast: I would like to use glib atomic rc box for refcounting.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    struct InstrumentBase *next; /* linked list in global registry */
+} InstrumentBase;
+
+struct ObservabilityCounter
+{
+    InstrumentBase base;
+    pthread_mutex_t lock;
+    CounterDataPoint *dataPoints;
+    int dataPointCount;
+    int dataPointCapacity;
```
cleithner-comcast: I don't think we should be using C arrays wherever possible. We can use lists.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    HistogramDataPoint *dataPoints;
+    int dataPointCount;
+    int dataPointCapacity;
+};
+
+/* ------------------------------------------------------------------ */
+/* Global registry                                                    */
+/* ------------------------------------------------------------------ */
+
+static pthread_mutex_t registryLock = PTHREAD_MUTEX_INITIALIZER;
```
cleithner-comcast: Use ERROR_CHECK

cleithner-comcast (changes requested):

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+
+        if (val == NULL)
+        {
+            break;
+        }
+
+        if (set.count >= cap)
+        {
+            cap *= 2;
+            set.pairs = realloc(set.pairs, sizeof(AttrPair) * cap);
```
rchowdcmcsa: no null check. There are a couple of other calls to malloc/calloc etc where we also don't null check the pointer. Maybe not a big deal but I think it's good practice.

core/src/observability/observability.h
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// ------------------------------ tabstop = 4 ----------------------------------
+
+#ifndef OBSERVABILITY_INIT_H
+#define OBSERVABILITY_INIT_H
```
rchowdcmcsa: nit: doesn't match actual file name

core/test/src/observabilityMetricsTest.c
```diff
@@ -0,0 +1,436 @@
...
+    assert_string_equal(
+        cJSON_GetStringValue(cJSON_GetObjectItem(cJSON_GetObjectItem(metrics, "multi.histogram"), "type")), "histogram");
+
+    /* Verify description and unit are included */
+    assert_string_equal(
+        cJSON_GetStringValue(cJSON_GetObjectItem(cJSON_GetObjectItem(metrics, "multi.gauge"), "unit")), "bytes");
+
+    cJSON_Delete(root);
+    free(json);
+    observabilityCounterRelease(c);
```
rchowdcmcsa, line 574: Should probably also add a test that tests observabilityCounterRelease(c) and whether or not the instrument is still present in observabilityDumpJson() after running it.

rchowdcmcsa (approved):

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    {
+        if (attrSetEqual(&counter->dataPoints[i].attrs, attrs))
+        {
+            return &counter->dataPoints[i];
+        }
+    }
+
+    if (counter->dataPointCount >= counter->dataPointCapacity)
+    {
+        counter->dataPointCapacity = counter->dataPointCapacity == 0 ? 4 : counter->dataPointCapacity * 2;
```
kfundecmcsa: `4` and `2` look like magic numbers - probably best to #define these

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -0,0 +1,835 @@
...
+    }
+
+    pthread_mutex_unlock(&registryLock);
+}
+
+/* ------------------------------------------------------------------ */
+/* Counter                                                            */
+/* ------------------------------------------------------------------ */
+
+static CounterDataPoint *counterFindOrAdd(ObservabilityCounter *counter, const AttrSet *attrs)
```
kfundecmcsa: "Add" is a bit overloaded in this file (`observabilityCounterAdd` adds to a value, `counterFindOrAdd` adds a slot), so it's a little tricky to understand from the function name, and there's no doc comment to fill in the gap. Could we rename to something like `counterGetOrCreateDataPoint` (same for the gauge/histogram twins) and add a brief comment noting that each unique attribute set gets its own lazily-allocated data point?

kfundecmcsa (changes requested):

core/src/observability/observability.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @return 0 on success, non-zero on failure
+ */
+int observabilityInit(void);
+
+/**
+ * Shut down the observability subsystem and release all instruments.
+ * Safe to call even if init was not called or failed.
+ */
+void observabilityShutdown(void);
```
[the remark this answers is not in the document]
rchowdcmcsa: I have opted to change the docstring to match the current behavior. But perhaps we want the opposite? @tleacmcsa

config/cmake/options.cmake
rchowdcmcsa: bcore_string_option always creates a preprocessor definition via target_compile_definitions, right? I think we just don't need that here, so it wasn't used. I don't know if our preference is to just use bcore_*option everywhere regardless though.

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: ERROR_CHECK?

cleithner-comcast: PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP

https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3.html

error check mutexes provide some layers of protection in alerting you to mutex misuse. Linux only, the NP stands for non-portable.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -37,13 +37,15 @@
...
 
+#include <glib.h>
 #include <pthread.h>
 #include <stdarg.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <stdlib.h>
 #include <string.h>
 
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
```
cleithner-comcast: We don't have to define this, it exists in linux headers. Glib also has G_N_ELEMENTS which does the same thing.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -82,65 +81,62 @@ static AttrSet attrSetFromVa(va_list ap)
...
 
-    for (int i = 0; i < a->count; i++)
+    while (nodeA != NULL && nodeB != NULL)
     {
-        if (strcmp(a->pairs[i].key, b->pairs[i].key) != 0 ||
-            strcmp(a->pairs[i].value, b->pairs[i].value) != 0)
+        AttrPair *pairA = nodeA->data;
+        AttrPair *pairB = nodeB->data;
+
+        if (strcmp(pairA->key, pairB->key) != 0 || strcmp(pairA->value, pairB->value) != 0)
```
cleithner-comcast: g_strcmp0 is NULL safe

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -167,9 +163,12 @@ typedef struct GaugeDataPoint
...
 
 /* Default bucket boundaries matching OpenTelemetry SDK defaults */
-static const double kHistogramBounds[] = {
-    0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000};
-static const int kNumBounds = sizeof(kHistogramBounds) / sizeof(kHistogramBounds[0]);
+static const double histogramBounds[] = {0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000};
+
+enum
+{
+    numBounds = ARRAY_SIZE(histogramBounds)
```
cleithner-comcast: an enum? Not sure I understand the purpose here.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -365,75 +335,76 @@ void observabilityCounterAddWithAttrs(ObservabilityCounter *counter, uint64_t va
...
-
-        free(counter->dataPoints);
-        free(counter->base.name);
-        free(counter->base.description);
-        free(counter->base.unit);
-        free(counter);
+        return;
     }
+
+    registryRemove(&counter->base);
```
cleithner-comcast: This looks like a logic bug.

core/src/observability/inmemory/observabilityInMemory.c
```diff
@@ -613,38 +581,41 @@ void observabilityHistogramRecordWithAttrs(ObservabilityHistogram *histogram, do
...
-
-        free(histogram->dataPoints);
-        free(histogram->base.name);
-        free(histogram->base.description);
-        free(histogram->base.unit);
-        free(histogram);
+        return;
     }
+
+    registryRemove(&histogram->base);
```
cleithner-comcast: Same here, another logic bug.

cleithner-comcast (changes requested):

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: fixed

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: fixed

core/src/observability/memory/observabilityMemory.c
rchowdcmcsa: done

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: done

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: kNumBounds was originally a static const int, which in C is not an integer constant expression, so using it as a struct array bound technically makes it a variable-length array, which clang tolerates as a GNU extension but warns about. The change to an enum was a Copilot suggested change that I just accepted to get rid of the warning.

We could just not care about the warning, or change it to something else.

config/cmake/options.cmake
cleithner-comcast: I think we should not be concerned with superfluous preprocessor definitions, they'd be compiled out if unreferenced. I suppose this is just a copilot thing, not deliberate. Using the bcore_* APIs is ideal because it creates a consistent interface of repeated operations.

core/src/observability/inmemory/observabilityInMemory.c
cleithner-comcast: Looked into this, very neat. constexpr looks like the proper way but requires C23+. This enum is fine. Another thing that should work is just using ARRAY_SIZE type macros since they expand to sizeof expressions which are compile-time computable. Either way, good with this.

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: fixed.

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: changed to use g_strcmp0

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: Yeah, Copilot went and added those invocations in all the `*Release` functions and I didn't catch it. Removed.

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: Fixed

core/src/observability/memory/observabilityMemory.c
```diff
@@ -0,0 +1,831 @@
...
+}
+
+ObservabilityCounter *observabilityCounterCreate(const char *name, const char *description, const char *unit)
+{
+    if (name == NULL || name[0] == '\0')
+    {
+        return NULL;
+    }
+
+    ObservabilityCounter *c = g_atomic_rc_box_new0(ObservabilityCounter);
```
kfundecmcsa, line 302: I don't see any corresponding usages of `g_atomic_rc_box_acquire` anywhere in the file (or via a public `observability*Acquire`) - doesn't this mean the refcount is always 1 for the lifetime of the object?

kfundecmcsa (changes requested):

core/src/observability/memory/observabilityMemory.c
kfundecmcsa, line 1: Since the CMake stuff was changed from `inmemory` to `memory` - do we want this file (and parent dir) to just be `memory` now?

core/src/observability/memory/observabilityMemory.c
rchowdcmcsa: Hmm, yes. The same situation existed in the original version of this PR. `refCount` initialized to 1, decremented in `*Release`, but there was no `*Acquire` or increment anywhere. I didn't really question it the first time I read over the code.

We could just replace this with `g_new0`/`g_free` but maybe we prefer the overhead of ref counting for future robustness or something like that.

core/src/observability/memory/observabilityMemory.c
rchowdcmcsa: I'll defer to others on that, don't feel too strongly.

tleacmcsa: yeah, easy refactor and keeps in sync.  i vote to do it.

rchowdcmcsa: done

core/src/observability/inmemory/observabilityInMemory.c
rchowdcmcsa: forgot to comment before -- done.

cleithner-comcast (approved):

core/src/observability/memory/observabilityMemory.c
cleithner-comcast: Refcounting overhead is basically free, we should be using it nearly everywhere but especially for interface types like these. You can never predict how they are going to be used completely nor whether requirements for them will change. The day someone needs to use one of these like a shared type, they aren't going to do the work to add refcounting properly, they're going to do some silly funky mutexing that will introduce another subtle maintenance bug 2 years later. Avoid all of that by implementing refcounting from the start.

kfundecmcsa: Fair enough

kfundecmcsa (approved):

core/src/observability/memory/observabilityMemory.c
cleithner-comcast: FWIW g_atomic_rc_box_acquire "just works" but it would be proper to have an abstraction (so we can change the implementation of refcounting under the hood and not break clients). We should do so.

cleithner-comcast (changes requested):

core/src/observability/memory/observabilityMemory.c
```diff
@@ -0,0 +1,831 @@
...
+/* ------------------------------------------------------------------ */
+/* Global registry                                                    */
+/* ------------------------------------------------------------------ */
+
+static pthread_mutex_t registryLock = PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP;
+static GSList *registry = NULL;
+
+static void registryAdd(InstrumentBase *inst)
+{
+    pthread_mutex_lock(&registryLock);
```
cleithner-comcast: I failed to mention that ERRORCHECK benefits require processing return codes. We have convenience macros for this in xhConcurrent's threadUtils.h. Use mutexLock, mutexUnlock, and LOCK_SCOPE where appropriate. Otherwise the program won't properly abort when mutexes are misused.

cleithner-comcast: @rdkcentral/copilot-comcast-developers replace usage of pthread_mutex_lock and pthread_mutex_unlock in this source file with mutexLock and mutexUnlock from <icConcurrent/threadUtils.h>

cleithner-comcast: Dangit, that was meant to be for copilot. Frustrating that's the first autocomplete.

cleithner-comcast: @Copilot replace usage of pthread_mutex_lock and pthread_mutex_unlock in this source file with mutexLock and mutexUnlock from <icConcurrent/threadUtils.h>

cleithner-comcast (approved):

rchowdcmcsa (approved):

PR #236, opened 2026-06-25
feat(sbmd): replace v3 YAML runtime with v4 JavaScript-native architecture

cleithner-comcast:
Replace the v3 YAML-based SBMD runtime (SbmdParser, SbmdSpec, SbmdScript, ScriptResult) with a v4 JavaScript-native architecture:

- SbmdDriver: manages per-driver QuickJS context lifecycle and GC roots
- SbmdDispatch: builds and queries handler dispatch tables from driver registrations
- SbmdLoader: loads and evaluates .sbmd.js spec files in QuickJS
- SbmdHandlerInvoker: marshals C++ context into JS args, invokes handlers, and unmarshals results
- SbmdResultExecutor: interprets result chain arrays into C++ side effects
- SbmdBundleLoader: assembles and loads the runtime JS bundle (renamed from SbmdUtilsLoader)
- SbmdRegistration: header-only types for driver self-registration

Remove the v3 YAML parser, v2/v3 JSON schemas, and dual-engine SbmdScriptImpl. Add v4 JSON schema for spec validation.

Includes comprehensive unit tests for all new components and updated MatterDevice to use the v4 dispatch-based driver model.

Also (unrelated to SBMD v4):

- Expose observability metrics (#235) via new `b_core_client_get_telemetry` public API and `getTelemetry` command in the reference app.


api/c/src/barton-core-client.c
```diff
@@ -191,6 +192,26 @@ BCoreStatus *b_core_client_get_status(BCoreClient *self)
...
+    if (json == NULL)
+    {
+        return NULL;
+    }
+
+    /* Transfer ownership to glib — observabilityDumpJson uses malloc,
+     * but the public API contract says g_free(). Copy into g_strdup
+     * and free the original. */
+    gchar *result = g_strdup(json);
+    free(json);
```
tleacmcsa, line 210: Yuck

core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.h
```diff
@@ -36,63 +36,55 @@ extern "C" {
...
-     *   // Create invoke response for command
-     *   return SbmdUtils.Response.invoke(0x0006, 0x0001);  // On command
-     * @endcode
+     * The bundle is assembled at build time from individual source files:
+     *   1. sbmd-namespace.js — creates the Sbmd namespace and _internal
+     *   2. sbmd-utf8.js      — adds Sbmd._internal.Utf8
+     *   3. sbmd-base64.js    — adds Sbmd.Base64
+     *   4. sbmd-tlv.js       — adds Sbmd.Tlv
+     *   5. sbmd-result.js    — adds Sbmd.result() builder
+     *   6. sbmd-cleanup.js   — removes Sbmd._internal
```
tleacmcsa: This module shouldnt know about these files, which are provided at build time.

tleacmcsa (commented):
OMG...

api/c/src/barton-core-client.c
```diff
@@ -191,6 +192,26 @@ BCoreStatus *b_core_client_get_status(BCoreClient *self)
     return convertDeviceServiceStatusToGObject(status);
 }
 
+gchar *b_core_client_get_telemetry(BCoreClient *self)
```
kfundecmcsa, line 195: The telemetry bits (`b_core_client_get_telemetry` + reference-app `getTelemetry` command) seem out of place for this PR, but probably should explain this addition in the commit message.

kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.h
tleacmcsa: done.

api/c/src/barton-core-client.c
tleacmcsa: no way around this for now since cJSON uses malloc... someone would have to strdup it into glib.  oh well.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.h
```diff
@@ -36,63 +36,49 @@ extern "C" {
...
-         * @param ctx The mquickjs context to load the utilities into
-         * @return true if the utilities were loaded successfully, false otherwise
+         * @param ctx The mquickjs context to load the bundles into
+         * @return true if all bundles were loaded successfully, false otherwise
          */
         static bool LoadBundle(JSContext *ctx);
 
         /**
-         * Check if the SBMD utilities bundle is available.
+         * Check if the SBMD bundles are available.
```
kfundecmcsa: What does "available" mean here? The implementation is a compile-time #if HAS_EMBEDDED_BUNDLE check that returns true in any normal build (which the existing comment concedes — "should always be true when properly built"). I also can't find any caller of IsAvailable() or GetSource() — both look like dead API.

Either delete them, or clarify the contract (e.g. "returns true if the embedded bundle was compiled in") and document why they exist if there's a planned use I'm not seeing.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+        JSValue args = JS_NewObject(ctx);
+
+        // GC-root args during construction: subsequent allocations (JS_NewString,
+        // JS_NewObject) may trigger mquickjs GC which would sweep the unrooted args.
+        JSGCRef argsRef {};
+        JS_AddGCRef(ctx, &argsRef);
+        argsRef.val = args;
+
+
+        JS_SetPropertyStr(ctx, args, "deviceUuid", JS_NewString(ctx, hctx.deviceUuid.c_str()));
```
kfundecmcsa: Lots of magic-string property names here that are actually the wire contract with the `.sbmd.js` handlers (`args.deviceUuid`, `args.attribute.clusterId`, `args.attribute.tlvBase64`, etc). A typo on the C++ side just shows up as undefined in JS, no build break. There's also real duplication: "clusterId" 5x, "tlvBase64" 3x, "handlerContext" 3x, plus "attribute" / "attributeId" / "commandId" each twice.

Could we at least pull these into #defines at the top of the file?

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+    std::optional<ParsedResult> SbmdHandlerInvoker::InvokeHandler(JSContext *ctx, JSValue handler, JSValue args)
+    {
+        if (JS_IsUndefined(handler))
+        {
+            icError("handler is undefined");
+            return std::nullopt;
+        }
+
+        if (JS_StackCheck(ctx, 3))
```
kfundecmcsa: The `3` here is hard to make sense of without scanning the next few lines to count the `JS_PushArg` calls. Could we give it a name (`constexpr int callArgs = 3`) or at least an inline `// args, handler, this`?

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+    void SbmdHandlerInvoker::ExecuteOps(const HandlerContext &hctx,
+                                        const std::vector<ResultOp> &ops,
+                                        const TransientDataSetter &transientSetter)
+    {
+        for (const auto &op : ops)
+        {
+            if (std::holds_alternative<ResultOp::UpdateResource>(op.data))
+            {
+                const auto &ur = std::get<ResultOp::UpdateResource>(op.data);
+                const char *epId = ur.endpoint.has_value() ? ur.endpoint->c_str() : hctx.endpointId.c_str();
```
kfundecmcsa, line 286: This took me a minute to follow. The two branches of this ternary are do very different things. One writes to the endpoint the handler explicitly named, the other silently redirects the write to the endpoint that triggered the handler. That's a meaningful behavioral fork hiding inside a one-liner, and it's not obvious from the expression which case is the default or why falling back to the trigger endpoint is the right choice. Worth a short comment so the next reader doesn't have to reverse-engineer the intent.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+extern "C" {
+#include <icLog/logging.h>
+#include <mquickjs/mquickjs.h>
+}
+
+namespace barton
+{
+    namespace
+    {
+        std::string GetExceptionString(JSContext *ctx)
```
kfundecmcsa: `GetExceptionString` is copy-pasted into the anonymous namespace of at least three files in this tree (here, `SbmdBundleLoader`, `MQuickJsRuntime`), identical body in each. MQuickJsRuntime already owns related exception plumbing (CheckAndClearPendingException), so promoting this to a public static there and deleting the local copies seems like the natural fix. Same pattern exists in the parallel `quickjs/` tree, so worth deciding whether each runtime gets its own or there's a shared helper.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+            }
+
+            JSCStringBuf buf;
+            const char *str = JS_ToCString(ctx, val, &buf);
+
+            return str ? std::string(str) : "";
+        }
+
+        /**
+         * Get a uint32 property from a JS object, or 0 if missing.
```
kfundecmcsa: Is "missing = 0" the right default here? There's already a `GetOptUint32Prop` right below for the "might be missing" case, so I'm wondering why some call sites pick this one. For things like `minSecs`, `clusterId`, or `profileVersion`, 0 seems like a valid value, so a typo'd or omitted key in the driver JS becomes indistinguishable from an intentional `0` - and for genuinely required fields (driverVersion, clusterId), shouldn't a missing key fail the load instead of silently becoming 0?

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+
+            return result;
+        }
+
+        /**
+         * Get the keys of a JS object by evaluating Object.keys().
+         * We store the object on a temporary global, evaluate Object.keys(),
+         * then clean up.
+         */
+        std::vector<std::string> GetObjectKeys(JSContext *ctx, JSValue obj)
```
kfundecmcsa: This function looked a bit odd to me so I dug into it. I did an experiment locally and it looks like we can simplify it to:

```cpp
JS_SetPropertyStr(ctx, JS_GetGlobalObject(ctx), "__sbmd_tmp", obj);
const char *script = "Object.keys(__sbmd_tmp)";
JSValue keysArr = JS_Eval(ctx, script, strlen(script), "<keys>", JS_EVAL_RETVAL);
JS_SetPropertyStr(ctx, JS_GetGlobalObject(ctx), "__sbmd_tmp", JS_UNDEFINED);
if (JS_IsException(keysArr)) {
    MQuickJsRuntime::CheckAndClearPendingException(ctx);
    return {};
}
return GetStringArray(ctx, keysArr);
```

`Object.keys()` already returns a JS array of strings, so we don't need to JSON.stringify it and parse the result back out — we can hand the array straight to the `GetStringArray` helper that's already defined a few lines up in this file. That deletes the hand-rolled JSON parser entirely (~50 lines → ~12), and as a side effect we stop having to worry about whether that parser handles things like `\"`, `\\`, `\n`, or `\uXXXX` escapes in key names correctly.

Ran the full SbmdLoaderTest suite against this proposed change and all 25 tests pass, so the alias / endpoint / handler key-enumeration paths all behave identically.

tleacmcsa: just noticed this pattern of utility methods using JS_Eval to do work. I wonder how expensive this is.  Feels like it might be costly in general.  I don't have an issue with your change, just wondering about the overall cost wherever we do it.

kfundecmcsa: ah good point, that's worth looking into

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+            icError("SbmdDriver injection left a pending exception: %s", exMsg.c_str());
+            return false;
+        }
+
+        icDebug("SbmdDriver capture function injected");
+        return true;
+    }
+
+    std::vector<std::pair<std::string, std::string>> SbmdLoader::ExtractConstants(JSContext *ctx,
+                                                                                     const char *source,
```
kfundecmcsa: nit: spacing is off here

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        return preamble;
+    }
+
+    int SbmdLoader::CountPreambleLines(const std::string &preamble)
+    {
+        return static_cast<int>(std::count(preamble.begin(), preamble.end(), '\n'));
+    }
+
+    std::unique_ptr<SbmdRegistration> SbmdLoader::LoadDriver(JSContext *ctx,
+                                                                   const std::string &filePath,
```
kfundecmcsa: nit: spacing is off

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        if (!ctx || !source || sourceLen == 0)
+        {
+            icError("Invalid arguments to LoadDriver");
+            return nullptr;
+        }
+
+        icDebug("Loading driver from %s (%zu bytes)", filePath.c_str(), sourceLen);
+
+        // Pass 1: Extract constants
+        auto constants = ExtractConstants(ctx, source, sourceLen);
```
kfundecmcsa, line 451: Is it intentional that `ExtractConstants` returns `{}` for both “no constants block” and error cases?  

The header documents empty as “no constants block,” but implementation also returns empty on parse/eval/rejection failures.  
Since this call site just continues, we can’t tell "no constants block" from extraction-failed. Should we split those outcomes?

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.h
```diff
@@ -0,0 +1,157 @@
...
+         * Scans for "constants:" or "constants :" followed by "{", then brace-matches
+         * to find the closing "}". Evaluates the block as "({...})" to get an object,
+         * then walks its properties to produce name=value pairs.
+         *
+         * @param ctx The mquickjs context
+         * @param source The file contents
+         * @param sourceLen Length of the file contents
+         * @return Vector of (name, value-as-JS-literal) pairs, empty if no constants block
+         */
+        static std::vector<std::pair<std::string, std::string>> ExtractConstants(JSContext *ctx,
```
kfundecmcsa: I might be reading this wrong, but are ExtractConstants / GenerateConstantsPreamble / CountPreambleLines meant to be public API, or are they only public so tests can call them directly?

They feel like internal LoadDriver details. If that’s the intent, should we keep them private and cover them through LoadDriver behavior instead?

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+               reg->name.c_str(),
+               filePath.c_str(),
+               reg->schemaVersion.c_str(),
+               reg->driverVersion);
+
+        return reg;
+    }
+
+    std::unique_ptr<SbmdRegistration> SbmdLoader::ExtractRegistration(JSContext *ctx,
+                                                                           const std::string &filePath)
```
kfundecmcsa: nit: spacing off

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        {
+            if (!ExtractDeviceHandlers(ctx, cmdHandlers, reg->commandHandlers))
+            {
+                icError("Failed to extract commandHandlers from %s", filePath.c_str());
+                return nullptr;
+            }
+        }
+
+        // Reset __sbmd_registration to null for the next driver
+        JS_SetPropertyStr(ctx, global, "__sbmd_registration", JS_NULL);
```
kfundecmcsa: This reset looks pretty citical for isolation between driver loads. Should we check the return from `JS_SetPropertyStr(...)` here? If this write fails, we could carry stale `__sbmd_registration` into the next driver load.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        if (!JS_IsUndefined(supplementsVal) && !JS_IsNull(supplementsVal))
+        {
+            handler.supplements = ExtractSupplements(ctx, supplementsVal);
+        }
+
+        return handler;
+    }
+
+    bool SbmdLoader::ExtractDeviceHandlers(JSContext *ctx,
+                                              JSValue handlersObj,
```
kfundecmcsa: nit: spacing off

core/deviceDrivers/matter/sbmd/mquickjs/SbmdResultExecutor.cpp
```diff
@@ -0,0 +1,363 @@
...
+                return "";
+            }
+
+            JSCStringBuf buf;
+            const char *str = JS_ToCString(ctx, val, &buf);
+
+            return str ? std::string(str) : "";
+        }
+
+        uint32_t GetUint32Prop(JSContext *ctx, JSValue obj, const char *name)
```
kfundecmcsa: These helper functions in the anonymous namespace are duplicated in SbmdLoader (`GetStringProp`, `GetUint32Prop`, `GetOptUint32Prop`, `GetOptUint16Prop`, `GetArrayLength`) - probably best to break thise out into an internal helper to get rid of the duplication.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdResultExecutor.cpp
```diff
@@ -0,0 +1,363 @@
...
+        {
+            ResultTerminal::Error data;
+            data.message = GetStringProp(ctx, termVal, "message");
+
+            return ResultTerminal{std::move(data)};
+        }
+        else if (opType == "sendCommand")
+        {
+            ResultTerminal::SendCommand data;
+            data.clusterId = GetUint32Prop(ctx, termVal, "clusterId");
```
kfundecmcsa: same issue as in SbmdHandlerInvoker - lots of magic-string property names that are duplicated a bunch. Probably best to #define these at the top - maybe even consider moving these to a common module if they're used across modules.

core/deviceDrivers/matter/sbmd/schema/CHANGELOG.md
```diff
@@ -1,23 +0,0 @@
-# SBMD Schema Changelog
```
kfundecmcsa, line 1: should we delete the changelog altogether? I imagine there will be changes down the road to the new version we will want to track. Maybe we can just have like

```
## v4.0

- Initial strictly javascript versioned schema
```

or something like that

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema-v4.0.json
```diff
@@ -0,0 +1,402 @@
...
+          { "type": "boolean" }
+        ]
+      }
+    },
+    "aliases": {
+      "type": "object",
+      "description": "Named references to Matter cluster attributes, events, or commands.",
+      "additionalProperties": { "$ref": "#/$defs/alias" }
+    },
+    "barton": { "$ref": "#/$defs/barton" },
```
kfundecmcsa: Could we add object-level descriptions for `barton` / `matter` / `reporting` (both at the top-level refs and in $defs)? The child fields are documented, but the objects themselves are currently undocumented.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.h
tleacmcsa: it should be available if the build succeeds.  Feel free to remove this method and make sure the build fails if it would end up in whatever state "not available" would be.

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <algorithm>
+
+extern "C" {
+#include <icLog/logging.h>
+}
+
+namespace barton
+{
+    void SbmdDispatchTable::Build(const std::unordered_map<std::string, SbmdAlias> &aliases,
+                                   const std::vector<SbmdDeviceHandler> &handlers)
```
kfundecmcsa: nit: spacing off

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
```diff
@@ -0,0 +1,160 @@
...
+     */
+    class SbmdDispatchTable
+    {
+    public:
+        /**
+         * Build a dispatch table from the registration's aliases and a handler vector.
+         *
+         * @param aliases The registration's alias map (name → SbmdAlias)
+         * @param handlers The device handler vector (attributeHandlers, eventHandlers, or commandHandlers)
+         * @param aliasElementGetter Function to extract the relevant element ID from an alias
```
kfundecmcsa: this isn't a parameter

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
tleacmcsa: Sure.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
tleacmcsa: i like the latter

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+            {
+                JS_DeleteGCRef(ctx, &pending.onResponseRef);
+                pending.onResponseRooted = false;
+            }
+
+            if (pending.onErrorRooted)
+            {
+                JS_DeleteGCRef(ctx, &pending.onErrorRef);
+                pending.onErrorRooted = false;
+            }
```
rchowdcmcsa: Shouldn't there also be a block like below in this spot? Or am I missing something.

```C++
if (pending.contextRooted)
{
    JS_DeleteGCRef(ctx, &pending.contextRef);
    pending.contextRooted = false;
}
```

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: might be better to have another parameter to pass in as the default if the property is missing

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+        if (!tlvBase64.empty())
+        {
+            JS_SetPropertyStr(ctx, response, "data", JS_NewString(ctx, tlvBase64.c_str()));
+        }
+        else
+        {
+            JS_SetPropertyStr(ctx, response, "data", JS_NULL);
+        }
+
+        JS_SetPropertyStr(ctx, args, "response", response);
```
cleithner-comcast: Attaching resposnse after populating response. This line needs to be the first line after initializing response because JS_NewString can cause it to be garbage collected.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+        JSGCRef argsRef {};
+        JS_AddGCRef(ctx, &argsRef);
+        argsRef.val = args;
+
+
+        JSValue attribute = JS_NewObject(ctx);
+        JS_SetPropertyStr(ctx, attribute, "clusterId", JS_NewUint32(ctx, clusterId));
+        JS_SetPropertyStr(ctx, attribute, "attributeId", JS_NewUint32(ctx, attributeId));
+        JS_SetPropertyStr(ctx, attribute, "value", JS_NewString(ctx, tlvBase64.c_str()));
+        JS_SetPropertyStr(ctx, args, "attribute", attribute);
```
cleithner-comcast: Attribue populated before being attached to args, JS_New* APIs can GC it.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+        if (matterCode >= 0)
+        {
+            JS_SetPropertyStr(ctx, error, "matterCode", JS_NewInt32(ctx, matterCode));
+        }
+        else
+        {
+            JS_SetPropertyStr(ctx, error, "matterCode", JS_NULL);
+        }
+
+        JS_SetPropertyStr(ctx, args, "error", error);
```
cleithner-comcast: error populated before attach to args, JS_New* APIs can GC it.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+
+        CompletePendingOperation(pendingId, true);
+        return;
+    }
+
+    if (std::holds_alternative<ResultTerminal::RequestCommand>(result.terminal.data))
+    {
+        const auto &cmd = std::get<ResultTerminal::RequestCommand>(result.terminal.data);
+
+        // Re-arm: release old GC roots, root new handlers
```
cleithner-comcast: This re-arm sequence (for deferred callback chains) acquires the context lock and re-roots pending.onResponseRooted and pending.onErrorRooted. These are rooted with cmd.onResponse and cmd.onError respectively. The problem is that JSValues were parsed in a different critical section. Meaning that prior to this scope acquiring the lock, there was a gap where cmd.onResponse and cmd.onError could have been garbage collected due to concurrent operations.

Additionally, cmd.context (the deferred handler's opaque context) is never re-rooted, so it is silently drop in deferred chains with more than 1 link.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
```diff
@@ -84,66 +144,206 @@ namespace barton
...
+        std::optional<uint8_t> ConvertModesToBitmask(const std::vector<std::string> &modes);
 
         /** Map of device ID to set of resource keys (endpointId:resourceId) for optional resources that failed
-         * configuration */
+         * prerequisite checks */
         std::map<std::string, std::set<std::string>> skippedOptionalResources;
 
+        /** Active deferred operations indexed by unique ID */
+        std::map<uint64_t, PendingOperation> pendingOperations;
+        uint64_t nextPendingId = 1;
```
cleithner-comcast, line 388: transientStore, skippedOptionalResources, pendingOperations, and nextPendingId are not thread safe but instead rely on the undocumented behavior that all Do* operations will be scheduled on the Matter thread via a Schedule* call. This is pretty par for the course for our drivers, but some documentation could be nice.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+
+        // GC-root args across AddSupplements (which allocates) and InvokeHandler
+        JSGCRef argsRef {};
+        JS_AddGCRef(ctx, &argsRef);
+        argsRef.val = args;
+
+
+        if (matterDevice)
+        {
+            SbmdHandlerInvoker::AddSupplements(ctx,
```
cleithner-comcast: AddSupplements makes calls to deviceService* APIs. This is being executed while the mquickjs runtime mutex is held, which can threaten a lock order bug such as a reentrant call into this driver.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+    };
+}
+
+void SpecBasedMatterDeviceDriver::SetTransientData(const std::string &deviceUuid,
+                                                   const std::string &key,
+                                                   const std::string &value,
+                                                   uint32_t ttlSecs)
+{
+    auto expiry = std::chrono::steady_clock::now() + std::chrono::seconds(ttlSecs);
+    transientStore[deviceUuid][key] = TransientEntry {value, expiry};
```
rchowdcmcsa, line 1964: Worth protecting `transientStore` with a mutex? I can't tell if this is thread-safe or not (can't tell if it's always accessed in the Matter thread context).

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+        {
+            icWarn("Command handler '%s' returned no result for cluster 0x%x command 0x%x",
+                   entry->handler->name.c_str(),
+                   clusterId,
+                   commandId);
+            continue;
+        }
+
+        // Execute ops (updateResource, setMetadata, etc.)
+        SbmdHandlerInvoker::ExecuteOps(hctx, result->ops, MakeTransientSetter(deviceId));
```
cleithner-comcast: We only inspect the result for an error terminal, but do not catch if an improper terminal (eg, ResultTerminal::SendCommand) is returned. We should log when an incorrect terminal is returned.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+        if (!result.has_value())
+        {
+            icWarn("Event handler '%s' returned no result for cluster 0x%x event 0x%x",
+                   entry->handler->name.c_str(),
+                   clusterId,
+                   eventId);
+            continue;
+        }
+
+        // Execute ops (updateResource, setMetadata, etc.)
```
cleithner-comcast: We only inspect the result for an error terminal, but do not catch if an improper terminal (eg, ResultTerminal::SendCommand) is returned. We should log when an incorrect terminal is returned.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+        if (!result.has_value())
+        {
+            icWarn("Attribute handler '%s' returned no result for cluster 0x%x attr 0x%x",
+                   entry->handler->name.c_str(),
+                   clusterId,
+                   attributeId);
+            continue;
+        }
+
+        // Execute ops (updateResource, setMetadata, etc.)
```
cleithner-comcast: We only inspect the result for an error terminal, but do not catch if an improper terminal (eg, ResultTerminal::SendCommand) is returned. We should log when an incorrect terminal is returned.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+
+        // GC-root args across AddSupplements (which allocates) and InvokeHandler
+        JSGCRef argsRef {};
+        JS_AddGCRef(ctx, &argsRef);
+        argsRef.val = args;
+
+
+        if (matterDevice)
+        {
+            SbmdHandlerInvoker::AddSupplements(ctx,
```
cleithner-comcast: AddSupplements makes calls to deviceService* APIs. This is being executed while the mquickjs runtime mutex is held, which can threaten a lock order bug such as a reentrant call into this driver.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
```
cleithner-comcast: AddSupplements makes calls to deviceService* APIs. This is being executed while the mquickjs runtime mutex is held, which can threaten a lock order bug such as a reentrant call into this driver.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
```diff
@@ -29,19 +29,73 @@
...
+
+        // GC-rooted deferred handlers (stored in JSGCRef for proper GC management)
+        JSGCRef onResponseRef {};
+        JSGCRef onErrorRef {};
+        bool onResponseRooted = false;
+        bool onErrorRooted = false;
+
+        // Match criteria (for requestCommand responses)
+        uint32_t clusterId = 0;
+        uint32_t responseCommandId = 0;
```
cleithner-comcast: These two members are written but never read. Matter ExchangeManager matches request command responses so I believe they can be removed.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+            *pending.executeResponse = strdup(cmd.successValue.c_str());
+        }
+
+        // The command was sent. Completion comes via the command's own promise.
+        // The parking promise remains pending until that resolves.
+        // For sendCommand in a chain, we complete the parking promise when the
+        // command completes, which happens via OnDone → promise.set_value(true).
+        // We need to wait for that promise and then complete ours.
+        // Actually, the tempPromises will be resolved when the command completes.
+        // We'll complete the parking promise as success since the command was accepted.
```
cleithner-comcast: AI offloaded its thinking context to this comment block. This needs rewriting.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
```diff
@@ -0,0 +1,1087 @@
...
+            icError("Failed to evaluate constants block: %s", GetExceptionString(ctx).c_str());
+            return constants;
+        }
+
+        // Get the keys and values
+        auto keys = GetObjectKeys(ctx, objVal);
+
+        for (const auto &key : keys)
+        {
+            JSValue val = JS_GetPropertyStr(ctx, objVal, key.c_str());
```
cleithner-comcast, line 350: values are serialized below but keys are now. Later, when we call SbmdLoader::GenerateConstantsPreamble, we construct "var key = value;\n" lines. Except, keys may not be valid identifiers. This leaves a gap for malformed statements or even timmy drop tables injections.

core/deviceDrivers/matter/sbmd/SbmdDriver.cpp
```diff
@@ -0,0 +1,265 @@
...
+{
+    SbmdDriver::SbmdDriver(std::unique_ptr<SbmdRegistration> registration, std::string source) :
+        registration(std::move(registration)), source(std::move(source))
+    {
+    }
+
+    SbmdDriver::~SbmdDriver()
+    {
+        // If still activated at destruction, the GC refs are leaked.
+        // This shouldn't happen in normal operation.
```
cleithner-comcast: It would behoove us to document the intended usage gap: Deactivate before delete.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+    {
+        icDebug("Empty TLV data for attribute 0x%x", attributeId);
+        return;
+    }
+
+    // Base64 encode the TLV data
+    size_t maxBase64Len = BASE64_ENCODED_LEN(tlvLen) + 1;
+    std::string tlvBase64(maxBase64Len, '\0');
+    uint16_t encoded = chip::Base64Encode(tlvBuf, static_cast<uint16_t>(tlvLen), tlvBase64.data());
+    tlvBase64.resize(encoded);
```
cleithner-comcast: TLV encoding block is copy/pasted 5 times in this file, should be in a helper.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+}
+
+TransientDataSetter SpecBasedMatterDeviceDriver::MakeTransientSetter(const std::string &deviceUuid)
+{
+    return [this, deviceUuid](const std::string &key, const std::string &value, uint32_t ttlSecs) {
+        SetTransientData(deviceUuid, key, value, ttlSecs);
+    };
+}
+
+void SpecBasedMatterDeviceDriver::HandleAttributeReport(const std::string &deviceId,
```
cleithner-comcast, line 2098: HandleAttributeReport, HandleEvent, HandleCommand can likely be templated.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -0,0 +1,527 @@
...
+{
+    JSValue SbmdHandlerInvoker::BuildBaseArgs(JSContext *ctx, const HandlerContext &hctx)
+    {
+        JSValue args = JS_NewObject(ctx);
+
+        // GC-root args during construction: subsequent allocations (JS_NewString,
+        // JS_NewObject) may trigger mquickjs GC which would sweep the unrooted args.
+        JSGCRef argsRef {};
+        JS_AddGCRef(ctx, &argsRef);
+        argsRef.val = args;
```
cleithner-comcast: We can create a trivial RAII ScopedGCRoot for this pattern.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+    {
+        std::lock_guard<std::mutex> lock(MQuickJsRuntime::GetMutex());
+        auto *ctx = MQuickJsRuntime::GetSharedContext();
+        JS_DeleteGCRef(ctx, &deferredRoots->onResponse);
+        JS_DeleteGCRef(ctx, &deferredRoots->onError);
+        JS_DeleteGCRef(ctx, &deferredRoots->context);
+    }
+}
+
+void SpecBasedMatterDeviceDriver::ExecuteTerminal(std::forward_list<std::promise<bool>> &promises,
```
cleithner-comcast, line 966: ExecuteTerminal and ContinueDefferredChain have a lot of duplication. We can likely consolidate.

core/deviceDrivers/matter/sbmd/scriptCommon/sbmd-result.js
```diff
@@ -0,0 +1,384 @@
...
+                        commandId: commandId,
+                        deferred: {}
+                    };
+
+                    if (tlv !== undefined && tlv !== null)
+                    {
+                        t.tlvBase64 = tlv;
+                    }
+
+                    if (opts !== undefined)
```
rchowdcmcsa: Should also be `if (opt !== undefined && opt !== null)`, I believe, similar to the above block.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
```diff
@@ -29,19 +29,73 @@
...
+
+        // Deferred handler options
+        JSGCRef contextRef {};
+        bool contextRooted = false;
+
+        // Context for handler invocation
+        HandlerContext handlerContext;
+
+        // Context for continuing the chain
+        MatterDevice *device = nullptr;
```
rchowdcmcsa: Risk of raw dangling pointer here I think. Would be better to just store the device UUID here and then use it to call GetDevice() as needed.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
-                    clusterId,
-                    device.GetDeviceId().c_str());
+            uint8_t resourceMode = *modeResult;
 
-            return false;
-        }
+            if (resource.execute.has_value())
+            {
+                resourceMode |= RESOURCE_MODE_EXECUTABLE;
+            }
```
rchowdcmcsa, line 627: Why put this check here? Is `*modeResult` not sufficient in determining whether or not a resource is executable? As written, doesn't this imply that the modes array might not declare that a resource is executable while the resource might in fact have an execute handler? Isn't that a problem? Shouldn't we be relying on one source of truth here?

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -566,129 +463,2011 @@ uint8_t SpecBasedMatterDeviceDriver::ConvertModesToBitmask(const std::vector<std
...
+
+        // The command was sent. Completion comes via the command's own promise.
+        // The parking promise remains pending until that resolves.
+        // For sendCommand in a chain, we complete the parking promise when the
+        // command completes, which happens via OnDone → promise.set_value(true).
+        // We need to wait for that promise and then complete ours.
+        // Actually, the tempPromises will be resolved when the command completes.
+        // We'll complete the parking promise as success since the command was accepted.
+        CompletePendingOperation(pendingId, true);
+        return;
```
rchowdcmcsa: This is just a really confusing comment. I don't understand what's being said here. It sounds unfinished and self-contradictory. I ran it through AI, and the AI was confused too.

Are we just trying to say that we're resolving the promise optimistically?

core/test/src/SbmdLoaderTest.cpp
```diff
@@ -0,0 +1,692 @@
...
+        auto constants = ExtractConstants(R"(
+            SbmdDriver({ constants: { A: "hello \"world\"", B: "back\\slash" } });
+        )");
+
+        ASSERT_EQ(constants.size(), 2u);
+        EXPECT_EQ(constants[0].first, "A");
+        EXPECT_EQ(constants[0].second, R"("hello \"world\"")");
+        EXPECT_EQ(constants[1].first, "B");
+        EXPECT_EQ(constants[1].second, R"("back\\slash")");
+    }
```
rchowdcmcsa, line 154: Worth adding a test to catch things like \n, \r, as well? ExtractConstantsControlCharacters() or something.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.h
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done - made it optional

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.h
kfundecmcsa: probably should be private, but left them as public for now to facilitate testing - may revisit at some point

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: made this into a function since it's done in several spots and checks for an exception

core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdResultExecutor.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdResultExecutor.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/schema/CHANGELOG.md
kfundecmcsa: restored the changelog

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema-v4.0.json
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: added

core/deviceDrivers/matter/sbmd/scriptCommon/sbmd-result.js
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.h
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: left this as-is but modified docs to be more clear about what's expected here.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: fixed

core/test/src/SbmdLoaderTest.cpp
rchowdcmcsa: added

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
```diff
@@ -0,0 +1,158 @@
...
+}
+
+namespace barton
+{
+    /**
+     * Priority level for handler matching.
+     */
+    enum class HandlerPriority
+    {
+        Specific, // Single alias → single (clusterId, elementId)
```
kfundecmcsa: `HandlerPriority::Specific` clashes with `specificTable` — same word, but they mean different things. The enum value is just `aliases.size() == 1` (single alias), while `specificTable` is keyed by a specific element — which could come from a single-alias or multi-alias handler. Confusing that they share the name.

`HandlerPriority::Specific` should be renamed to `Single` (or similar) so it's not stepping on `specificTable`.

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * A dispatch table that maps (clusterId, elementId) to a priority-sorted
+     * list of handler entries. Also maintains a wildcard table keyed by
+     * clusterId only.
+     */
+    class SbmdDispatchTable
+    {
+    public:
+        /**
+         * Build a dispatch table from the registration's aliases and a handler vector.
```
kfundecmcsa: This doc just restates the signature — doesn't say what `Build()` actually does: clears the tables, resolves each handler's aliases to `(clusterId, elementId)`, and sorts them into `specificTable` vs `wildcardTable` (unknown aliases dropped with a warning). Worth spelling that out.

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
```diff
@@ -0,0 +1,192 @@
...
+                {
+                    elementId = alias.commandId;
+                }
+
+                if (!elementId.has_value())
+                {
+                    // Wildcard — no specific element ID, matches all in this cluster
+                    wildcardTable[alias.clusterId].push_back(
+                        DispatchEntry{entry.handler, HandlerPriority::Wildcard});
+                    continue;
```
kfundecmcsa: Minor: the `continue` reads a little awkwardly here since it's really just a two-way split. An `if (!elementId) { ... } else { ... }` would make "wildcard vs specific" more obvious at a glance.

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+        // Sort each entry list by priority (Specific < Multi < Wildcard)
+        for (auto &[key, entries] : specificTable)
+        {
+            std::stable_sort(entries.begin(), entries.end(), [](const DispatchEntry &a, const DispatchEntry &b) {
+                return static_cast<int>(a.priority) < static_cast<int>(b.priority);
+            });
+        }
+
+        for (auto &[clusterId, entries] : wildcardTable)
```
kfundecmcsa: This second loop is a no-op — every entry in `wildcardTable` is pushed with `HandlerPriority::Wildcard` (it's the only insertion point), so a bucket's priorities are all identical and there's nothing to sort. Should be dropped.

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
```diff
@@ -0,0 +1,192 @@
...
+                        DispatchEntry{entry.handler, HandlerPriority::Wildcard});
+                    continue;
+                }
+
+                DispatchKey key{alias.clusterId, elementId.value()};
+                specificTable[key].push_back(entry);
+            }
+        }
+
+        // Sort each entry list by priority (Specific < Multi < Wildcard)
```
kfundecmcsa: `Wildcard` entries won't happen in the `specificTable` - that should be dropped from this comment

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
```diff
@@ -0,0 +1,158 @@
...
+         */
+        std::vector<const DispatchEntry *> Lookup(uint32_t clusterId, uint32_t elementId) const;
+
+        /**
+         * Clear all entries.
+         */
+        void Clear();
+
+        /**
+         * Get the number of specific+multi entries (for diagnostics).
```
kfundecmcsa: "(for diagnostics)" is limiting — that's just how it's used right now, not what it does. Might want it for other things later. Let's drop that bit (also in the `GetWildcardEntryCount`)

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
kfundecmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDispatch.h
kfundecmcsa: done

rchowdcmcsa (approved):

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -58,154 +59,111 @@ extern bool deviceServiceSetMetadata(const char *uri, const char *value);
 
 namespace barton
 {
-    JSValue SbmdHandlerInvoker::BuildBaseArgs(JSContext *ctx, const HandlerContext &hctx)
+    SafeJSValue SbmdHandlerInvoker::BuildBaseArgs(JSContext *ctx, const HandlerContext &hctx)
     {
-        JSValue args = JS_NewObject(ctx);
+        // Root args for the whole build: subsequent allocations (JS_NewString, JS_NewObject)
+        // may trigger the mquickjs moving GC, which would otherwise sweep the unrooted args.
+        SafeJSValue args {ctx, JS_NewObject(ctx)};
```
tleacmcsa, line 66: It seems like construction could be a lot nicer for callers: SafeJSValue args(ctx), right?

cleithner-comcast: I'm on the fence. Yes it would be more convenient but it obfuscates allocation. `SafeJSValue args {ctx};` doesn't indicate an allocation has occurred and the only place the wrapper does allocate is self-documenting via the AddObject call. A reader could think `SafeJSValue args {ctx}` is an empty wrapper with a context but it would have a legitimately rooted jsvalue.

I'm not opposed to it I just kinda like the explicit client opt-in to allocation. Will change if you feel strongly though.

kfundecmcsa (approved):

PR #237, opened 2026-06-29
fix: cleith/dev/sbmd-v4-3-bugfixes

cleithner-comcast:


core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -148,24 +154,52 @@ bool SpecBasedMatterDeviceDriver::AddDevice(std::unique_ptr<MatterDevice> device
                                       chip::ClusterId clusterId,
                                       chip::CommandId commandId,
                                       chip::TLV::TLVReader &reader) {
-        // Encode TLV as base64 for HandleCommand
-        uint8_t tlvBuf[1024];
-        chip::TLV::TLVWriter writer;
-        writer.Init(tlvBuf, sizeof(tlvBuf));
-
+        // Encode the incoming command's TLV payload as base64 for HandleCommand.
+        // The buffer must hold the entire command fields struct; WebRTC SDP
```
tleacmcsa: I'm not sure I like specific use cases discussed in generic/common places like this.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -148,24 +154,52 @@ bool SpecBasedMatterDeviceDriver::AddDevice(std::unique_ptr<MatterDevice> device
...
+            CHIP_ERROR err = writer.CopyElement(chip::TLV::AnonymousTag(), readerCopy);
 
-            if (tlvLen > 0)
+            if (err == CHIP_ERROR_BUFFER_TOO_SMALL || err == CHIP_ERROR_NO_MEMORY)
             {
-                size_t maxBase64Len = BASE64_ENCODED_LEN(tlvLen) + 1;
-                tlvBase64.resize(maxBase64Len, '\0');
-                uint16_t encoded = chip::Base64Encode(tlvBuf, static_cast<uint16_t>(tlvLen), tlvBase64.data());
-                tlvBase64.resize(encoded);
+                bufSize *= 2;
```
tleacmcsa, line 176: we could go ahead and use observability stuff for this since its already merged.  We should certainly keep an eye on buffers like this.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -783,9 +843,47 @@ void SpecBasedMatterDeviceDriver::HandleResourceOp(std::forward_list<std::promis
...
+                    hasTempRoots = true;
+                }
+            }
+            else if (std::holds_alternative<ResultTerminal::ReadAttribute>(result->terminal.data))
+            {
+                const auto &ra = std::get<ResultTerminal::ReadAttribute>(result->terminal.data);
+
+                if (!JS_IsUndefined(ra.onResponse) || !JS_IsUndefined(ra.onError) || !JS_IsUndefined(ra.context))
+                {
+                    JS_AddGCRef(ctx, &tempOnResponse);
```
tleacmcsa: duplication here looks unnecessary

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
cleithner-comcast: Agree. Do you want to start adding them incrementally or should we do a full scoped audit and exploration later?

tleacmcsa: I'm fine adding it afterward. We need this PR to keep going.

core/deviceDrivers/matter/sbmd/SbmdDriver.h
```diff
@@ -139,8 +139,10 @@ namespace barton
 
         /**
          * Add a GC root for a handler JSValue if it is not JS_UNDEFINED.
+         * On success, rootedOut is set to the GC-tracked storage location (updated by the moving GC);
+         * callers must invoke through that location, not the raw handler copy.
          */
-        void RootIfValid(JSContext *ctx, JSValue &handler);
+        void RootIfValid(JSContext *ctx, JSValue &handler, JSValue *&rootedOut);
```
tleacmcsa: how about using a template so you dont need to pass each handler element individually?  Something like:
```
       template <typename T>
        void RootIfValid(JSContext *ctx, T &handlerEntry)
        {
            RootIfValid(ctx, handlerEntry.handler, handlerEntry.rooted);
        }
```
and

`RootIfValid(ctx, handler);`

core/deviceDrivers/matter/sbmd/SbmdRegistration.h
```diff
@@ -74,8 +74,14 @@ namespace barton
      */
     struct SbmdResourceHandler
```
tleacmcsa: Should these be classes now, with a reset method to set handler to JS_UNDEFINED and rooted to nullptr (and whatever else is needed going forward)?  Wouldnt need the template approach suggested above and could just use base class.

kfundecmcsa (approved):
Approved - Although I don’t have deep hands-on experience with GC/rooting safety, so I’m deferring to your expertise on that part. I reviewed this mainly from a design/readability perspective, and it looks good.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
```diff
@@ -288,6 +288,22 @@ bool MQuickJsRuntime::CheckAndClearPendingException(JSContext *ctx, std::string
...
+                JSCStringBuf buf;
+                const char *propStr = JS_ToCString(ctx, propVal, &buf);
+                if (propStr)
+                {
+                    exMsg += " | ";
+                    exMsg += prop;
+                    exMsg += "=";
+                    exMsg += propStr;
+                }
+            }
```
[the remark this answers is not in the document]
cleithner-comcast: Incorrect

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1068,38 +1177,43 @@ void SpecBasedMatterDeviceDriver::ExecuteRequestCommand(std::forward_list<std::p
...
+    // until after the emplace — moving the PendingOperation into the map would
+    // otherwise invalidate the registered address. Note also that JS_AddGCRef
+    // resets ref->val, so values must be assigned AFTER the call.
+    pending.onResponseRooted = !JS_IsUndefined(cmd.onResponse);
+    pending.onErrorRooted = !JS_IsUndefined(cmd.onError);
+    pending.contextRooted = !JS_IsUndefined(cmd.context);
+
+    auto [it, inserted] = pendingOperations.emplace(pendingId, std::move(pending));
+    PendingOperation &stored = it->second;
+
```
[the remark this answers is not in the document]
cleithner-comcast: Incorrect

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
cleithner-comcast: Same, reworded.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/SbmdDriver.h
cleithner-comcast: Went with the class path, lemme know what you think

core/deviceDrivers/matter/sbmd/SbmdRegistration.h
cleithner-comcast: Done.

cleithner-comcast:
> Approved - Although I don’t have deep hands-on experience with GC/rooting safety, so I’m deferring to your expertise on that part. I reviewed this mainly from a design/readability perspective, and it looks good.

The way to think about it is this:

All mquickjs types (JSValue, for example) are backed by a vary particular heap memory management technique. This includes the loaded specs, which lodge routine text data into heap memory and hands out refrence addresses to those symbols. Whenever you call into mquickjs for non-trivial context operations (example, executing a script) the library lazily audits its heap memory. This means it will look to clean data up, move data around, pack it, etc. When this happens, those old addresses that it handed out are now undefined. They may point to the same symbols, they may not. GCRefs are basically a mechanism on top that updates mquickjs types (JSValues) in step with the garbage collection. Meaning if I use a GCRef on my JSValue that points to my javascript function, the original JSValue could become stale but the GCRef isn't.

This commit does a lot of work to make sure important stuff does not have critical sections where they can be garbage collected without a GCRef while they are still needed. Additionally, static duration things like handler functions get a pointer to the GCRef of the original handler, which basically means they now can invoke that handler even if it gets moved around in the heap. The GCRef model has some pitfalls (such as you MUST add it to the context before setting the val member because adding the ref to a context clears the val member) so we do need to be pretty particular about usage and also have a good idea of how the library works going forward.

kfundecmcsa:
> > Approved - Although I don’t have deep hands-on experience with GC/rooting safety, so I’m deferring to your expertise on that part. I reviewed this mainly from a design/readability perspective, and it looks good.
> 
> The way to think about it is this:
> 
> All mquickjs types (JSValue, for example) are backed by a vary particular heap memory management technique. This includes the loaded specs, which lodge routine text data into heap memory and hands out refrence addresses to those symbols. Whenever you call into mquickjs for non-trivial context operations (example, executing a script) the library lazily audits its heap memory. This means it will look to clean data up, move data around, pack it, etc. When this happens, those old addresses that it handed out are now undefined. They may point to the same symbols, they may not. GCRefs are basically a mechanism on top that updates mquickjs types (JSValues) in step with the garbage collection. Meaning if I use a GCRef on my JSValue that points to my javascript function, the original JSValue could become stale but the GCRef isn't.
> 
> This commit does a lot of work to make sure important stuff does not have critical sections where they can be garbage collected without a GCRef while they are still needed. Additionally, static duration things like handler functions get a pointer to the GCRef of the original handler, which basically means they now can invoke that handler even if it gets moved around in the heap. The GCRef model has some pitfalls (such as you MUST add it to the context before setting the val member because adding the ref to a context clears the val member) so we do need to be pretty particular about usage and also have a good idea of how the library works going forward.

Very good explanation, thanks!

core/deviceDrivers/matter/sbmd/SbmdRegistration.h
```diff
@@ -67,15 +67,38 @@ namespace barton
...
+        }
+    };
+
     /**
      * A resource handler declaration (seed, read, write, or execute).
      * For simple declarations (just a function), only handler is set.
      * For object declarations, supplements and handler are both set.
      */
-    struct SbmdResourceHandler
+    struct SbmdResourceHandler : SbmdHandlerBase
```
tleacmcsa: This isnt needed anymore, eh?  Maybe just rename SbmdHandlerBase to SbmdHandler and use it directly?  I dont feel strongly.

tleacmcsa (approved):

rchowdcmcsa (approved):

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1068,38 +1191,43 @@ void SpecBasedMatterDeviceDriver::ExecuteRequestCommand(std::forward_list<std::p
...
-        if (!JS_IsUndefined(cmd.onResponse))
+        if (stored.onResponseRooted)
         {
-            JS_AddGCRef(ctx, &pending.onResponseRef);
-            pending.onResponseRef.val = cmd.onResponse;
-
-            pending.onResponseRooted = true;
+            JS_AddGCRef(ctx, &stored.onResponseRef);
+            stored.onResponseRef.val = cmd.onResponse;
         }
```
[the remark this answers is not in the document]
cleithner-comcast: This was a real bug, I have had to address this with some shared_ptr composition in https://github.com/rdkcentral/BartonCore/pull/237/commits/637e56c94b2c5afb697948579f6de7aa5896f21a

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -2071,16 +2199,20 @@ void SpecBasedMatterDeviceDriver::HandleAttributeReport(const std::string &devic
...
     hctx.deviceUuid = deviceId;
     hctx.endpointId = std::to_string(endpointId);
-    // TODO: populate clusterFeatureMaps from MatterDevice
 
     auto matterDevice = GetDevice(deviceId);
 
+    if (matterDevice)
+    {
+        hctx.clusterFeatureMaps = matterDevice->GetCachedClusterFeatureMaps();
+    }
```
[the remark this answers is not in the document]
cleithner-comcast: Not concerned with deep copy at this time, these maps are tiny.

core/deviceDrivers/matter/sbmd/SbmdRegistration.h
cleithner-comcast: Done in https://github.com/rdkcentral/BartonCore/pull/237/commits/3b4cbb16bc803929049057c4335bd99d85a26f60

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -783,9 +856,48 @@ void SpecBasedMatterDeviceDriver::HandleResourceOp(std::forward_list<std::promis
...
+
+            if (deferredOnResponse != nullptr &&
+                (!JS_IsUndefined(*deferredOnResponse) || !JS_IsUndefined(*deferredOnError) ||
+                 !JS_IsUndefined(*deferredContext)))
+            {
+                JS_AddGCRef(ctx, &tempOnResponse);
+                JS_AddGCRef(ctx, &tempOnError);
+                JS_AddGCRef(ctx, &tempContext);
+                tempOnResponse.val = *deferredOnResponse;
+                tempOnError.val = *deferredOnError;
```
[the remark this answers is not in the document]
cleithner-comcast: This was a real bug, I have had to address this with some shared_ptr composition in https://github.com/rdkcentral/BartonCore/commit/637e56c94b2c5afb697948579f6de7aa5896f21a

tleacmcsa (approved):

kfundecmcsa (approved):

PR #239, opened 2026-07-08
fix(matter): populate VID/PID/flow in local commissioning payload

tleacmcsa:
When opening a commissioning window for the local node (node 0), the generated QR code and manual setup code reported zeros for vendor id, product id, and commissioning flow because SetupPayload was never populated with them.

Populate vendor id and product id from the DeviceInstanceInfoProvider (the same source the SDK uses for the DNS-SD TXT record) and add a new optional Barton property, barton.matter.commissioningFlow, to source the commissioning flow, since the SDK provides no hook for it. This is done only on the node-0 path, which advertises our own identity.

The remote path opens a commissioning window on another device and must report that device's values; leave those three fields as zeros for now with a TODO, which assumes the standard flow that permits zeros.

Refs: BARTON-398

### DIFF
--- a/api/c/public/barton-core-properties.h
+++ b/api/c/public/barton-core-properties.h
@@ -200,6 +200,18 @@ G_BEGIN_DECLS
     B_CORE_BARTON_PREFIX B_CORE_MATTER_PREFIX "setupPasscode"
 
 /**
+ * MATTER_COMMISSIONING_FLOW: (value "barton.matter.commissioningFlow")
+ *
+ * The commissioning flow advertised in the Matter device's onboarding payload
+ * (QR code and manual setup code). Valid values match the Matter Commissioning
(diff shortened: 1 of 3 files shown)


cleithner-comcast (dismissed):
Can we add integration tests for these three modes? I think we should be able to with matterjs using test credentials.

tleacmcsa:
> Can we add integration tests for these three modes? I think we should be able to with matterjs using test credentials.

No, not really. The infrastructure to actually test the modes involves DCL, web browsers, etc.

cleithner-comcast:
> > Can we add integration tests for these three modes? I think we should be able to with matterjs using test credentials.
> 
> No, not really. The infrastructure to actually test the modes involves DCL, web browsers, etc.

ack.

cleithner-comcast (approved):

PR #242, opened 2026-07-13
feat(matter): add gated verbose Matter message tracing

cleithner-comcast:
Add BCORE_MATTER_MESSAGE_TRACING (default OFF), a developer-only build option that logs every inbound/outbound Matter message as a JSON record (hex payload + fully-decoded TLV) inline via ChipLog.

- config/cmake/options.cmake: new bcore_option -> BARTON_CONFIG_MATTER_MESSAGE_TRACING
- Matter.cpp: register a JsonBackend on init and unregister on stop, guarded by the flag
- third_party/CMakeLists.txt: forward -t to build-matter.sh when the option is ON so a single build configures both the SDK and core
- build-matter.sh / barton-library build.sh: -t enables the tracing GN args (matter_enable_tracing_support + payload hex/decode)
- barton-library BUILD.gn: gate the //src/tracing/json dep on matter_enable_tracing_support
- docs/DEBUGGING.md: document the feature

### DIFF
--- a/build-matter.sh
+++ b/build-matter.sh
@@ -26,11 +26,12 @@
 set -e
 
 help() {
-    echo "Usage: $0 [-h] [-c <dir>] [-s] [-a]"
+    echo "Usage: $0 [-h] [-c <dir>] [-s] [-a] [-t]"
     echo "  -h  Display help"
     echo "  -c  Matter configuration directory"
     echo "  -s  Build with stack smash protection"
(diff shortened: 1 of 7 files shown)


tleacmcsa (approved):
Say "Developer feature" one more time... I dare you.

rchowdcmcsa (approved):

PR #243, opened 2026-07-14
fix(install): avoid LSAN hang in stub generation

tleacmcsa:
The post-install pygobject-stubs helper imports the ASAN-instrumented BartonCore library and exits immediately. In ASAN builds, LSAN teardown can hang in that short-lived import path, which leaves make install stuck after typelib installation.

Disable leak detection for the stub-generation helper while keeping the ASAN preload in place so the instrumented library still loads. This matches the existing GIR-generation environment and lets the install-time stub generation step exit cleanly.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -429,6 +429,12 @@ if (BCORE_GEN_GIR)
         elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
             set(ASAN_LD_PRELOAD "LD_PRELOAD=$(gcc -print-file-name=libasan.so)")
         endif()
+
+        # The pygobject-stubs generator imports the instrumented BartonCore
+        # library and exits immediately. In ASAN builds, LSAN teardown can hang
+        # during that short-lived import-only path, so disable leak detection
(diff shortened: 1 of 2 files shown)


core/CMakeLists.txt
```diff
@@ -429,6 +429,12 @@ if (BCORE_GEN_GIR)
         elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
             set(ASAN_LD_PRELOAD "LD_PRELOAD=$(gcc -print-file-name=libasan.so)")
         endif()
+
+        # The pygobject-stubs generator imports the instrumented BartonCore
+        # library and exits immediately. In ASAN builds, LSAN teardown can hang
+        # during that short-lived import-only path, so disable leak detection
+        # for this post-install helper just as we already do for GIR generation.
```
cleithner-comcast, line 436: Is there an issue describing this behavior that we can link to for this?

cleithner-comcast (approved):

core/CMakeLists.txt
tleacmcsa: No, I just hit it this morning and got this fix.

PR #244, opened 2026-07-14
fix(sbmd): harden mquickjs GC safety

tleacmcsa:
mquickjs uses a moving, compacting GC. Raw JSValue handles can go stale when allocations occur between property reads, stack growth, or argument construction. This change roots transient and retained values to prevent stale-handle and use-after-move behavior.

Changes:
- Root intermediate exception/property/array/result values with SafeJSValue in runtime, loader, util, invoker, and parser paths.
- Root handler and handlerContext before allocation-heavy operations.
- Split create-and-attach patterns in SafeJSValue helpers where argument evaluation order could pass stale parents under moving GC.
- Refine SbmdDriver teardown: detach held roots only when shared context is absent during runtime shutdown.
- Add a Date constructor compatibility shim for generated stdlib tables that reference js_date_constructor.
- Update SBMD tests to use rooted function accessors and root intermediate JS values in assertions.
- Enabled DEBUG_GC in mquickjs Docker build to help expose problematic JSValue usage.
- Bump mquickjs Docker pin and container version for reproducible builds aligned with these fixes.


core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
```diff
@@ -221,51 +222,57 @@ bool MQuickJsRuntime::CheckAndClearPendingException(JSContext *ctx, std::string
...
 
     if (!hasException)
     {
         return false;
     }
 
+    // mquickjs uses a moving/compacting GC: every JS allocation (including each JS_GetPropertyStr /
+    // JS_ToCString below) can relocate live objects, invalidating any raw JSValue held in a C
+    // local. Root the exception so it stays valid across the multiple property reads used to build
+    // the diagnostic message; otherwise the second and later reads dereference a stale pointer.
```
cleithner-comcast: Do we need this documentation everywhere? Doc slop

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsStdlib.c
```diff
@@ -105,6 +105,34 @@ static JSValue js_gc(JSContext *ctx, JSValue *this_val, int argc, JSValue *argv)
     return JS_UNDEFINED;
 }
 
+/*
+ * Date constructor compatibility shim.
+ *
+ * Newer generated mqjs_stdlib.h references js_date_constructor, but some
+ * mquickjs builds do not export an internal implementation symbol. Provide
+ * the constructor here so the generated stdlib table can always bind it.
```
cleithner-comcast: what?

cleithner-comcast (commented):
The comments are in excess, I think.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
tleacmcsa: fixed

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsStdlib.c
tleacmcsa: fixed.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsStdlib.c
```diff
@@ -105,6 +105,31 @@ static JSValue js_gc(JSContext *ctx, JSValue *this_val, int argc, JSValue *argv)
     return JS_UNDEFINED;
 }
 
+/*
+ * Compatibility function for generated mqjs_stdlib.h entries that reference
+ * js_date_constructor. Accepts an optional epoch-ms argument.
```
cleithner-comcast, line 110: What's the story here? Why do we define this and not mquickjs?

rchowdcmcsa (approved):

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsStdlib.c
tleacmcsa: yeah, its weird.  An mquickjs update we got removed this symbol, but its still used by mqjs_stdlib.c.  Something is fishy and I will look into it.

tleacmcsa: Its a funkiness in what they did to mquickjs.  I think it is a mistake on their end, but this symbol needs to be defined somewhere, either here or we patch it into mquickjs.  I bet they fix this upstream at some point.  We have other implementations in this file of stuff you would think would be in mquickjs, so I dont see great solutions.

cleithner-comcast: I would rather patch because otherwise we run into the issue we had in RDKB where we update mquickjs_git.bb and anything fixed on an older version of barton just breaks. Alternatively, we need to start versioning the mquickjs recipe (which maybe we need to do anyway) AND clients need to create a PREFERRED_VERSION coupling for it.

tleacmcsa: Oh, nevermind.  I see why its done this way.  Like all functions implemented in MQuickJsStdlib.c, these are the platform specific implementations required to support various aspects of the runtime.  Getting the current date, something this js_date_constructor does, requires platform support that mquickjs doesnt provide.  They are a strict "C only without dependencies" library.  We have to do this here.

cleithner-comcast: Okay, that makes sense. However, this creates a lock-step versioning problem. When we upgrade mquickjs and it introduces a new symbol, older versions of barton will have linker errors to it. Much like how clients already have to do PREFERRED_VERSION for matter, barton, and bartoncommon, we now need them to ALSO do one for mquickjs (and we need to version mquickjs in meta-rdk-iot).

tleacmcsa: Yeah.  I think we do need to version mquickjs recipes now so that we can manage that.  I'll deal with it in the meta-rdk-iot changes.

cleithner-comcast (approved):

PR #245, opened 2026-07-14
feat: add SBMD runtime observability instrumentation

rchowdcmcsa:
Instruments the SBMDv4 mquickjs runtime with OTel-compatible metrics:

**JS runtime** (`MQuickJsRuntime`): heap used/free/peak/arena gauges and histogram, mutex wait histogram, JS exception counter, background idle sampler (period: `BCORE_SBMD_METRICS_SAMPLE_PERIOD_MS`, default 30 s). `MQuickJsRuntime::AcquireMutex()` centralizes mutex acquisition and records wait time on every call. `MQuickJsRuntime` is a pure static class (constructor deleted, `inline static` members, C++17).

**GC** (new mquickjs patch `0003`): `sbmd.js.gc.count` counter, `sbmd.js.gc.duration_ms` histogram (via `JS_SetGCCallback`), and `sbmd.js.gc_roots` gauge (via `JS_GetGCRootCount`) for leak detection.

**Per-handler** (`SbmdHandlerInvoker`): duration and heap-delta histograms, outcome counter (`success`/`exception`/`timeout`/`stack_overflow`/`error`), all attributed by driver/op_type/resource_id. Timeout detection via `MQuickJsRuntime::WasTimedOut()` distinguishes script deadline expiry from handler exceptions.

**Driver loading** (`SbmdFactory`): load failure counter (by reason), duration and heap-delta histograms, registered driver count gauge. `RegisterDrivers()` is idempotent — returns immediately if drivers are already loaded. `MQuickJsRuntime::Initialize()` is called from `Matter::Matter()` so Initialize and Shutdown share the same lifecycle scope.

**Deferred ops** (`SpecBasedMatterDeviceDriver`): in-flight gauge, duration and depth histograms, timeout and depth-exceeded counters. `PendingOperation` carries an `OperationContext` (driver stem, op type, resource ID, start time) so attribution is consistent across the entire deferred chain.

Each metrics class initializes its handles in its constructor. Recording methods and constructors are `public`; instrument handles are `private`. No `MetricsRegistry`, no explicit `InitializeMetrics()`/`ShutdownMetrics()` lifecycle — handles persist for the process lifetime.

Set `BCORE_SBMD_METRICS=OFF` to compile out all metrics infrastructure: headers provide empty inline stubs, no sampler thread is created, zero overhead. `JS_SetGCCallback` call sites remain guarded with `#ifdef BARTON_CONFIG_SBMD_METRICS` since the patched symbols require conditional use; all other call-site guards have been removed (stubs absorb the calls when metrics are off).

GTest unit tests and integration tests cover all new metric paths. Integration tests for `sbmd.deferred.timeout` and `sbmd.deferred.max_depth` are planned as follow-on work (tasks 9.6 and 9.7).

Refs: BARTON-362


openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,134 @@
...
+- Changes to `core/src/observability/` infrastructure
+- Implicit or explicit GC tracking (follow-on)
+- `deviceId` attribute (follow-on)
+- Production/fleet identifier attributes
+
+## Decisions
+
+### Decision 1: Centralized `SbmdMetrics` class
+
+All metric handles (histograms, gauges, counters) live as static members of a new `SbmdMetrics` class in `core/deviceDrivers/matter/sbmd/SbmdMetrics.h/.cpp`. Recording functions (`RecordHandlerInvocation`, `RecordDriverLoad`, `RecordDeferredOpComplete`, etc.) are called from the existing code paths.
```
cleithner-comcast: I'm not sure I understand the point of this class. It reads like it's meant to be a garbage bag for static duration metric handles, which I'm not really in favor of. Good separation of concerns is allowing every module to instrument itself where appropriate. Tracing is the only thing that traverses frames but the observability APIs already support thread local retrieval of trace contexts so as to minimize polluting frame argument lists.

Is this class supposed to provide anything else besides just holding onto the handles?

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,134 @@
...
+
+**Why not create instruments inline at each call site?** Instrument creation has overhead and the observability spec requires a handle to persist for the lifetime of measurements. Centralizing in `SbmdMetrics` mirrors the `MQuickJsRuntime` singleton pattern already established in this subsystem.
+
+### Decision 2: `SbmdHandlerInvoker::InvokeHandler` signature extension
+
+`InvokeHandler` is extended with two optional parameters: `const char *driverName = nullptr, const char *opType = nullptr`. All existing callers compile unchanged (default nulls suppress attribution). Callers that know both values (`HandleResourceOp`, `HandleDeferredCommandResponse`, `HandleDeferredCommandError`, `ContinueDeferredChain`) pass them through.
```
cleithner-comcast: For what purpose? Is this to add attributes to metrics? I want to avoid modifying function signatures for the purposes of adding metric data. The reason for this is because it can become and ever growing list as the requirements for the telemetry changes over time.

Typical API behvaior is you demand exactly the arguments you need in order to complete a routine and a private context for any behaviors dictated by external responsibilities. For example, an API that takes a callback usually also takes an opaque context for that callback. It doesn't need the opaque context to complet its routine, but the client dictated callback may need arbitrary data to complete its callback.

I'm not necessarily saying an opaque context is the correct solution here, but I wouldn't be surprised if something like a generalized operation context that persists for the duration of one or more operations is.

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,134 @@
...
+**What is measured around `JS_Call`:**
+- `steady_clock::now()` before and after → invocation duration histogram
+- `JS_GetMemoryUsage` before and after → heap delta histogram
+- Exception / timeout / stack overflow distinction → outcome counter
+
+### Decision 3: Periodic heap sampler as a background thread in `MQuickJsRuntime`
+
+A `std::thread` is started at the end of `MQuickJsRuntime::Initialize()` and stopped in `Shutdown()`. The thread calls `SbmdMetrics::ForceSnapshot()` in a loop, sleeping for `BCORE_SBMD_METRICS_SAMPLE_PERIOD_MS` (default 30000 ms, configurable at CMake time) between calls. `ForceSnapshot()` handles all mutex management internally.
+
+**Why a background thread rather than sampling only on invocations?** Post-invocation sampling misses idle-state heap trends and slow leaks between bursts of activity. Pool utilization between invocations is a meaningful signal.
```
cleithner-comcast: The main issue with this is traffic is probably low for most installations so the capture will be 80% idle time, 20% activity time (making the numbers up, but you get the idea). I agree that only capturing during activity misses the baseline but I think only capturing periodically likely misses usage during activity most of the time.

Ideally, I would want manual captures during operations AND an idle background capture that can be tickled to reschedule. This basically would have non-busy installations lean on the idle thread capture most of the time, whereas the busy ones would rarely need to rely on the background capture because it will have activity all the time.

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,134 @@
...
+  ├── JS_NewContext(memBuffer, memSize)
+  ├── SbmdBundleLoader + SbmdLoader injection
+  └── starts periodicSamplerThread
+         │
+         └── loop:
+               SbmdMetrics::ForceSnapshot()  ← acquires mutex internally
+               sleep(SAMPLE_PERIOD_MS)
+```
+
+**Thread safety:** `ForceSnapshot()` acquires `MQuickJsRuntime::GetMutex()` before calling `JS_GetMemoryUsage`, the same mutex held during all JS execution. This prevents concurrent access to the JSContext.
```
cleithner-comcast: This single lock model is so messy. We gotta look to rework this at some point I think. At the very least a main loop model.

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Decision 4: `PendingOperation` extended with `startTime`
```
cleithner-comcast: Same comment earlier applies to any "extend foo" entries in here. Extending argument lists or existing classes may not be correct compared to a separate class to contextualize activity.

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: > It reads like it's meant to be a garbage bag for static duration metric handles, which I'm not really in favor of.

Yeah, for the most part, that is what it ended up as. The ForceSnapshot() thing is the only other real piece of non-trivial logic in there, but (if we keep it), it can just go somewhere else. I nixed this class.

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: Yeah, that was the purpose, and good call. I should've considered that.

Rather than add each individual attribute, we can propose just one function signature change that adds an opaque context. One of AI's suggestions was an ambient context set via RAII guard so that we don't have to change the function signature at all, but sounded a lot more error-prone and difficult to test to me; just thought I'd bring that up too.

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: Went ahead and amended the artifacts with your suggestion.

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: the aforementioned change (adding an opaque context) is in the current patch

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: removed

cleithner-comcast (dismissed):
If I might also suggest a gauge for GC root entries on the mquickjs list. I would like to know if we are "leaking" root entries over time from misuse of [Safe]JSValue. I am not sure the best way to do this, may require a patch to expose the list size.

openspec/changes/sbmd-observability-instrumentation/proposal.md
```diff
@@ -0,0 +1,46 @@
+## Why
+
+The SBMDv4 JavaScript runtime (mquickjs) runs inside a fixed 1 MB arena with a single shared JS context, making it a natural resource bottleneck. Today, nothing is measured: memory pressure, handler latency, GC activity, deferred operation health, and driver load cost are all invisible. Instrumenting the SBMD runtime with the existing observability API gives operators and developers the data needed to diagnose performance issues, detect failures, and understand system health.
+
+## What Changes
+
+- **Distributed metric instrumentation**: metric handles are owned by the module that records them. `MQuickJsRuntime` gains heap metrics and `ForceSnapshot()`; `SbmdHandlerInvoker` gains invocation and outcome metrics; `SbmdFactory` gains driver load metrics; `SpecBasedMatterDeviceDriver` gains deferred operation metrics.
```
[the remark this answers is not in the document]
rchowdcmcsa: will update PR description later.

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/specs/sbmd-runtime-observability/spec.md
```diff
@@ -0,0 +1,180 @@
...
+#### Scenario: Arena size recorded at initialization
+- **WHEN** `MQuickJsRuntime::Initialize(N)` is called
+- **THEN** the `sbmd.js.heap.arena_bytes` gauge records the value `N`
+
+#### Scenario: Free bytes reflects remaining capacity
+- **WHEN** a heap snapshot is taken (via `InvokeHandler` or `ForceSnapshot()`)
+- **THEN** `sbmd.js.heap.free_bytes` gauge records the current `usage.free_size` value (gap between heap top and stack bottom)
+
+### Requirement: Peak heap watermark gauge
+The SBMD runtime SHALL maintain a gauge named `sbmd.js.heap.peak_bytes` recording the highest value of `usage.heap_used` observed since `MQuickJsRuntime::Initialize()`. (`JS_MEMUSAGE_WALK_HEAP` is not set, so `usage.heap_free_blocks` is always 0; `heap_used` is used directly as the high-water mark of heap expansion.) The gauge SHALL be updated with each heap snapshot.
```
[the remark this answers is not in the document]
rchowdcmcsa: outdated PR description

rchowdcmcsa:
> If I might also suggest a gauge for GC root entries on the mquickjs list. I would like to know if we are "leaking" root entries over time from misuse of [Safe]JSValue. I am not sure the best way to do this, may require a patch to expose the list size.

Will explore this and figure out what we might be able to do.

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,160 @@
...
+- `sbmd.js.mutex.wait_ms` is measured in `HandleResourceOp` (`SpecBasedMatterDeviceDriver`), `sbmd.js.heap.*` pool health gauges are recorded from `SbmdHandlerInvoker::InvokeHandler`, and `sbmd.js.exception` is recorded from `SbmdLoader` and `SbmdBundleLoader`. All three sets of handles live in `MQuickJsRuntime`, which exposes `static void RecordMutexWait(double ms)`, `static void RecordJsException(const char *phase, const char *driver)`, and `static void RecordHeapSnapshot(JSContext *ctx)` (requires caller to hold the JS mutex).
+
+**Why not create instruments inline at each call site?** Instrument creation has overhead and the observability spec requires a handle to persist for the lifetime of measurements. Handles are created once in `InitializeMetrics()` and persist until `ShutdownMetrics()`.
+
+### Decision 2: `SbmdHandlerInvoker::InvokeHandler` invocation context parameter
+
+`InvokeHandler` is extended with a single optional parameter: `const SbmdOperationContext *opCtx = nullptr`. `SbmdOperationContext` is a plain struct defined in `SbmdHandlerInvoker.h`:
+
+```cpp
+struct SbmdOperationContext {
```
tleacmcsa: Maybe just OperationContext?

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,160 @@
...
+    const char *driverName = nullptr;
+    const char *opType = nullptr;   // originating op type (e.g., "write", "execute")
```
tleacmcsa: This gives us the operation type, but we should also track the related resource I think.

rchowdcmcsa: So like, `const char *resourceName`?

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,160 @@
...
+- **Deferred ops** (`HandleDeferredCommandResponse`, `HandleDeferredCommandError`, `ContinueDeferredChain`): caller passes `&pending.operationCtx` — the same object initialized when the `PendingOperation` was created and alive for the entire deferred chain.
+
+**Why a struct rather than individual parameters?** A single stable context slot avoids an ever-growing signature as requirements evolve. New fields are added to the struct without touching `InvokeHandler`'s signature or any call site. The context also carries natural operation-scope lifetime semantics: for deferred operations it is the same object from `ExecuteRequestCommand` through `CompletePendingOperation`, enabling `CompletePendingOperation` to read `startTime` from the same context that was active during every `InvokeHandler` call in the chain.
+
+**What is measured around `JS_Call`:**
+- `steady_clock::now()` before and after → invocation duration histogram
+- `JS_GetMemoryUsage` before and after → heap delta histogram
+- Exception / timeout / stack overflow distinction → outcome counter
+
+### Decision 3: Hybrid sampling — in-activity captures + tickleable idle thread
```
tleacmcsa, line 77: Since the sampling will sometimes be periodic when the timer expires and sometimes be driven by activity, wont this skew histograms and maybe other metrics since the sample period is random?

cleithner-comcast: I'm not sure, but in my mind activity bumps out the idle sample so your data is weighted towards which ever happens more often: idle or active. We maybe will need to adjust this implementation once we have real data.

openspec/changes/sbmd-observability-instrumentation/design.md
```diff
@@ -0,0 +1,160 @@
...
+│
+├── SbmdFactory.cpp                          ← driver load metrics
+└── SpecBasedMatterDeviceDriver.cpp          ← deferred op metrics; PendingOperation.operationCtx (SbmdOperationContext)
+```
+
+## Risks / Trade-offs
+
+- **In-activity heap snapshot overhead** → `RecordHeapSnapshot` is called at the end of every `InvokeHandler` invocation while the mutex is held. It reads the same `JS_GetMemoryUsage` struct already captured for the heap delta calculation — no additional JS engine work. Overhead is negligible.
+
+- **`JS_GetMemoryUsage` overhead inside `InvokeHandler`** → Called twice per invocation (before + after `JS_Call`). `JS_MEMUSAGE_WALK_HEAP` is not set — the call is an O(1) struct read with no heap walk. As a consequence, `usage.heap_free_blocks` is always 0 (per the mquickjs patch implementation); `sbmd.js.heap.free_bytes` therefore maps to `usage.free_size` (the gap between heap top and stack bottom), not to `heap_free_blocks`. Overhead is negligible.
```
tleacmcsa: We should probably define JS_MEMUSAGE_WALK_HEAP in our dev/docker build of mquickjs so we get actual numbers here.  No idea how much it would impact timing data.  Maybe we should measure time it takes to do that walk too, at least temporarily so we know if its an issue or not.

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
rchowdcmcsa: Maybe a gauge would work better than a histogram for this specifically.

cleithner-comcast: Traditionally, you upload all your data. Guages are cheaper than histograms, but the final data set uploaded is lossy because guages collection is often fixed interval. If you don't need high fidelity of information, you can rely on statistical analysis of uploaded guages (i.e. you put together a histogram of guages server-side). However, if you need high fidelity sampling, you instead collect a histogram locally.

In our case we aren't uploading this stuff right now so we can do either. But if we want to get statistical data on guages, we need some post processing of the dumped json (which we will probably want to do anyway). For this first stage, just running the report through copilot to produce something clean would probably be okay.

rchowdcmcsa: The main reason I suggested it is I feel like because the non-uniform sampling means that the histogram data will read like, e.g., "50% of the time that we checked, the heap was..." A gauge would be lossier but easier to interpret, I'd think.

If either is fine, we could just keep the histograms and make a note in the documentation that the data is "activity-weighted" or something (I'm not really sure how to phrase it).

rchowdcmcsa:
> > If I might also suggest a gauge for GC root entries on the mquickjs list. I would like to know if we are "leaking" root entries over time from misuse of [Safe]JSValue. I am not sure the best way to do this, may require a patch to expose the list size.
> 
> Will explore this and figure out what we might be able to do.

Two ways we could go about doing this I think.

1. Instrument `SafeJSValue` directly and count how many instances are alive at any moment (directly measures the concern of leaked `SafeJSValue`s).
2. Patch mquickjs to expose its internal GC root list size and measure the engine's own bookkeeping directly (would catch anything that touches `JS_AddGCRef` directly).

@cleithner-comcast

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
cleithner-comcast: Yeah I was thinking of this and I think I'm more interested in growing gc ref entries than I am specifically our class. Since the cause of leaking gc refs could be misuse of JSValue OR accidentally using JSValue directly or something else, I think measuring the class directly potentially doesn't capture other problems.

tleacmcsa:
> > > If I might also suggest a gauge for GC root entries on the mquickjs list. I would like to know if we are "leaking" root entries over time from misuse of [Safe]JSValue. I am not sure the best way to do this, may require a patch to expose the list size.
> > 
> > 
> > Will explore this and figure out what we might be able to do.
> 
> Two ways we could go about doing this I think.
> 
> 1. Instrument `SafeJSValue` directly and count how many instances are alive at any moment (directly measures the concern of leaked `SafeJSValue`s).
> 2. Patch mquickjs to expose its internal GC root list size and measure the engine's own bookkeeping directly (would catch anything that touches `JS_AddGCRef` directly).
> 
> @cleithner-comcast

1. That is a cool idea and worth a look!
2. We should not fear patching mquickjs for this (in fact, I already did -- see BartonCore/docker/patches/mquickjs/0001-add-memory-usage-api.patch)

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
cleithner-comcast: I replied to the wrong comment, I meant this for https://github.com/rdkcentral/BartonCore/pull/245#issuecomment-4983272821

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: I'll investigate this during implementation. Put it under "open questions" for now.

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
rchowdcmcsa: np, I understood

On the discussion here: I'm thinking of keeping it as a histogram, but adding a note about the non-uniform sampling under risks/tradeoffs for now (which I did in the latest patch that I pushed)

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: done

openspec/changes/sbmd-observability-instrumentation/design.md
rchowdcmcsa: done (added as resourceId)

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
tleacmcsa: I started this conversation, but will leave it to @cleithner-comcast to close when he is ready.

openspec/changes/sbmd-observability-instrumentation/design.md
tleacmcsa: Ok, ill close this for now but let's not forget about that option since I think it will be needed to get the full picture.

tleacmcsa (dismissed):

openspec/changes/sbmd-observability-instrumentation/proposal.md
```diff
@@ -0,0 +1,47 @@
+## Why
+
+The SBMDv4 JavaScript runtime (mquickjs) runs inside a fixed 1 MB arena with a single shared JS context, making it a natural resource bottleneck. Today, nothing is measured: memory pressure, handler latency, GC activity, deferred operation health, and driver load cost are all invisible. Instrumenting the SBMD runtime with the existing observability API gives operators and developers the data needed to diagnose performance issues, detect failures, and understand system health.
```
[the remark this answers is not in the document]
tleacmcsa: I agree that this needs some update.

openspec/changes/archive/2026-07-30-sbmd-observability-instrumentation/design.md
cleithner-comcast: I think either is feasable so I'm good with either approach and I think we can be incremental with these metrics anyway.

example-door-lock-sbmd.js
```diff
@@ -0,0 +1,175 @@
...
+
+// ===========================================================================
+// Driver Metadata
+// ===========================================================================
+
+SbmdDriver.define({
+  schemaVersion: "4.0",
+  driverVersion: "1.0",
+  name: "Door Lock",
+
```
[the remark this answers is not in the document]
rchowdcmcsa: Copilot acted out of turn. Fixing.

openspec/changes/sbmd-matter-critical-events/proposal.md
```diff
@@ -0,0 +1,36 @@
+## Why
+
+Matter Critical events (Informational or Critical priority) are delivered with higher reliability in congested mesh networks than attribute reports. Without event handlers, the three highest-priority sensor and lock drivers miss state updates that Matter devices are explicitly sending for reliability — defeating the purpose of the Critical event mechanism.
+
+## What Changes
+
+- **Door Lock driver** (`door-lock.sbmd.js`): Add event handlers for `DoorLockAlarm` and `LockOperation`. Introduce three new resources: `jammed`, `tampered`, and `invalidCodeEntryLimit` (all pre-defined in `commonDeviceDefs.h`). Bump `driverVersion` to 2.
+- **Contact Sensor driver** (`contact-sensor.sbmd.js`): Add event handler for `BooleanState.StateChange`. Bump `driverVersion` to 2.
+- **Water Leak Detector driver** (`water-leak-detector.sbmd.js`): Add event handler for `BooleanState.StateChange`. Bump `driverVersion` to 2.
```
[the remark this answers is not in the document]
rchowdcmcsa: Copilot acted out of turn. Fixing.

.vscode/settings.json
```diff
@@ -166,5 +166,7 @@
     "python.testing.cwd": "${workspaceFolder}",
     "python.testing.pytestArgs": ["testing"],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+
+    "chat.promptFiles": true
 }
```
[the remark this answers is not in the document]
rchowdcmcsa: Copilot acted out of turn. Fixing.

config/cmake/options.cmake
```diff
@@ -281,6 +281,11 @@ bcore_int_option(NAME BCORE_SBMD_SCRIPT_TIMEOUT_MS
                DESCRIPTION "Maximum execution time in milliseconds for SBMD mapper scripts."
                VALUE 5000)
 
+bcore_int_option(NAME BCORE_SBMD_METRICS_SAMPLE_PERIOD_MS
```
tleacmcsa: There should be a way to completely opt out of metrics gathering so that no resources are used ... everything would be a no-op.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -159,6 +165,78 @@ namespace barton
          */
         static std::chrono::steady_clock::time_point GetDeadline();
 
+        /**
+         * Mark that the script interrupt handler fired (deadline exceeded).
+         *
+         * Called exclusively from ScriptInterruptHandler when it returns 1.
```
tleacmcsa: I'm not a fan of code like this stating who the clients are and under what conditions they interact with this class.  Ideally code should define its interfaces with any relevant rules of use without these details.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
tleacmcsa, line 79: OMG.  Everything is static.  Is this what you were talking about the other day?  That is a crappy way to make a singleton.  I'm gonna puke.  AI did this.  ill create a ticket to address it.

tleacmcsa: https://ccp.sys.comcast.net/browse/BARTON-399

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -159,6 +165,78 @@ namespace barton
...
+         * Must not be called from any other context.
+         */
+        static void RecordInterrupt();
```
tleacmcsa: Why is this part of the public interface if its called internally?

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -159,6 +165,78 @@ namespace barton
...
+        // ----------------------------------------------------------------
+        // Observability
+        // ----------------------------------------------------------------
+
+        /**
+         * Initialize all metric handles owned by MQuickJsRuntime.
+         * Must be called before MQuickJsRuntime::Initialize() so that
+         * sbmd.js.exception is live for init-phase exceptions.
+         */
+        static void InitializeMetrics();
```
tleacmcsa: Should these observability parts be in some other class that is injected to MQuickJsRuntime if the feature is enabled?  Would that help it be testable without exposing these internals as public methods?

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: fixed

config/cmake/options.cmake
rchowdcmcsa: added something to accomplish this.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: I missed this -- will fix this and a few other instances of the same in an upcoming patch, but feel free to still review since it's just comment changes.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -167,6 +184,15 @@ namespace barton
...
+        static std::atomic<bool> scriptInterruptFired;
+
+        // jsContextReady: set true at end of Initialize(), cleared at start of
+        // Shutdown(). Used by MQuickJsRuntimeMetrics as a lock-free early-exit
+        // guard in ForceSnapshot().
+        static std::atomic<bool> jsContextReady;
+
+        // Called exclusively from ScriptInterruptHandler when it returns 1.
+        static void RecordInterrupt();
     };
```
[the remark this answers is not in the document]
rchowdcmcsa: fixed.

core/deviceDrivers/matter/sbmd/metrics/MetricsRegistry.h
```diff
@@ -0,0 +1,78 @@
...
+#include <functional>
+#include <vector>
+
+namespace barton
+{
+    /**
+     * Lightweight registry for SBMD metric lifecycle callbacks.
+     *
+     * Each metrics class self-registers at static-initialization time via
+     * registerProvider(); Matter.cpp calls initializeAll() / shutdownAll()
```
tleacmcsa: This module should not dictate/describe who the client(s) are.

core/deviceDrivers/matter/sbmd/metrics/MetricsRegistry.h
```diff
@@ -0,0 +1,78 @@
...
+            std::function<void()> shutdown;
+        };
+
+        static MetricsRegistry &instance();
+
+        /** Called from each metrics class's .cpp at file scope — before main(). */
+        static void registerProvider(Provider p);
+
+        /** Initialize all registered providers in registration order. */
+        static void initializeAll();
```
tleacmcsa: Is this really needed?  Cant providers initialize themselves?  Maybe shutdown autonomously as well?

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
```diff
@@ -0,0 +1,157 @@
...
+    public:
+        static void InitializeMetrics();
+        static void ShutdownMetrics();
+
+        /**
+         * Start the background idle heap sampler.
+         * No-op if BARTON_CONFIG_SBMD_METRICS_SAMPLE_PERIOD_MS <= 0.
+         * Called by MQuickJsRuntime::Initialize() after jsContextReady = true.
+         */
+        static void StartSampler();
```
tleacmcsa: I thought we agreed that we didn't like this singleton model (everything public static).  Does this even need to be a singleton?  Shouldnt this class (and others in the metrics directory) be with the thing it monitors?  That way MQuickJsRuntime just calls something like "monitor.Start();" or whatever, instead of MQuickJsRuntimeMetrics::Start().

core/deviceDrivers/matter/sbmd/metrics/MetricsRegistry.h
rchowdcmcsa: True, AI gets overzealous with its explanations and I miss it sometimes...will fix this and similar instances.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: We did agree on that, but you had created BARTON-399 to address this being a singleton model so I thought it was out of scope for this. If you'd rather I address it here, I can.

tleacmcsa: That ticket was to fix the exiting MQuickJsRuntime mess.  We should not continue that pattern with new code.

core/deviceDrivers/matter/sbmd/metrics/MetricsRegistry.h
rchowdcmcsa: I think we could drop the class, have the inits happen at static-init time, and let the preexisting observabilityShutdown() handle the shutdown for everything.

This class was something we came up with in an async discussion, but I can't quite remember what problem it was meant to solve now.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: Oh right, I understand what you're saying. I brainfarted on which file I was even looking at.
Yeah, I can fix this.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
```diff
@@ -137,82 +131,127 @@ bool MQuickJsRuntime::Initialize(size_t memorySize)
...
+    rt.initialized = true;
 
     // Install the script execution timeout interrupt handler
-    JS_SetInterruptHandler(ctx, ScriptInterruptHandler);
+    JS_SetInterruptHandler(rt.ctx, ScriptInterruptHandler);
     icDebug("Script execution interrupt handler installed");
 
+    // Install the GC instrumentation callback (passes the metrics instance as opaque)
+    JS_SetGCCallback(rt.ctx, &MQuickJsRuntimeMetrics::GCCallback, &rt.metrics);
+
```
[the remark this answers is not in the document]
tleacmcsa: @rchowdcmcsa please make sure it builds with the feature on and off.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
```diff
@@ -0,0 +1,156 @@
...
+
+namespace barton
+{
+    /**
+     * Observability metrics for the mquickjs JS runtime: heap pool health,
+     * GC activity, mutex contention, and JS exceptions.
+     */
+    class MQuickJsRuntimeMetrics
+    {
+        friend class MQuickJsRuntime;
```
tleacmcsa: Having public/private sections instead of friend is better in my opinion.  Allows for unit tests, etc.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
tleacmcsa: It looks like you partially converted this to a more standard singleton, but there are still static methods.  Isn't it half-and-half now?

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
```diff
@@ -91,32 +83,34 @@ namespace barton
 
 bool MQuickJsRuntime::Initialize(size_t memorySize)
 {
-    if (initialized)
+    auto &rt = GetInstance();
```
tleacmcsa: Lots of changes here due to the partial singleton fix.  These all go away once either the old/bad or new/full singleton model is implemented.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
```diff
@@ -0,0 +1,156 @@
...
+         */
+        void RecordJsException(const char *phase, const char *driver);
+
+        /**
+         * GC instrumentation callback — registered with JS_SetGCCallback with
+         * this instance as opaque. Called by the mquickjs engine at GC start
+         * (isEnd == 0) and end (isEnd == 1).
+         */
+        static void GCCallback(JSContext *ctx, int isEnd, void *opaque) noexcept;
+        void InitInstruments();
```
tleacmcsa: Why isnt this the constructor so it doesnt have to be called everywhere?

tleacmcsa (changes requested):
partial review.  Awaiting updates to continue due to thrash.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
rchowdcmcsa: If you want, I could just go all the way and move away from the singleton pattern completely and knock out the other ticket here too. It feels weird to leave it in this halfway state.
Alternatively I could make `metrics` and so on static and leave it for the other ticket, but yeah.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: good call, fixed.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
rchowdcmcsa: Made it all static again.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: fixed

config/cmake/options.cmake
```diff
@@ -281,6 +285,11 @@ bcore_int_option(NAME BCORE_SBMD_SCRIPT_TIMEOUT_MS
...
```
cleithner-comcast: Consider BCORE_SBMD_METRICS_HEAP_SAMPLE_PERIOD_MS to make room for other sampling in the future.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+                std::unique_lock<std::mutex> lock(samplerCvMutex);
+
+                while (!samplerShouldStop.load(std::memory_order_relaxed))
+                {
+                    auto deadline =
+                        clock::now() + std::chrono::milliseconds(BARTON_CONFIG_SBMD_METRICS_SAMPLE_PERIOD_MS);
+                    auto lastTickle = tickleSeq.load(std::memory_order_relaxed);
+
+                    while (true)
```
cleithner-comcast: This thread lambda could use some small documentation. The logic is not trivially clear and it took me a good minute to sort it out. Just some comments explaining intention or steps could prevent maintenance bugs.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
```diff
@@ -0,0 +1,226 @@
...
+                        {
+                            lock.unlock();
+                            ForceSnapshot();
+                            lock.lock();
+                            break;
+                        }
+
+                        auto curTickle = tickleSeq.load(std::memory_order_relaxed);
+
+                        if (curTickle != lastTickle)
```
cleithner-comcast: This is always a tautology in the happy path. Either an external client calls ForceSnapshot which tickles, or they tickle, or this thread calls ForceSnapshot which tickles. The only time this logic is diffrerent is if the runtime context is not ready, in which case this loop is never going to break out without intervention. I think you should review this logic and have copilot do the same.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -159,14 +161,76 @@ namespace barton
...
+        MQuickJsRuntime() = delete;
+
+        inline static uint8_t *memBuffer = nullptr;
+        inline static size_t memSize = 0;
+        inline static JSContext *ctx = nullptr;
+        inline static std::mutex mutex;
+        inline static bool initialized = false;
+        inline static size_t peakHeapUsed = 0;
+        inline static std::chrono::steady_clock::time_point deadline;
+        inline static std::atomic<bool> scriptInterruptFired {false};
```
cleithner-comcast: curious usage of inline

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -178,31 +182,86 @@ namespace barton
         if (JS_StackCheck(ctx, 3)) // args, handler, this
         {
             icError("stack overflow before handler call");
+#ifdef BARTON_CONFIG_SBMD_METRICS
```
cleithner-comcast: I don't think call sites should have ifdefs. Everywhere should agnostically call into the metrics and the implementation should just do nothing when not compiled in.

I looked ahead and there's quite a bit of ifdef in here. Try to ensure nothing outside of the metric classes reference the compile-time option. The runtime context installing the gc hook callback is a fine exception.

core/src/subsystems/matter/Matter.cpp
```diff
@@ -503,6 +508,10 @@ bool Matter::Stop()
 
     state.store(State::stopped);
 
+    // Stop the SBMD JS runtime -- joins the background heap sampler thread before
+    // the metric handles it uses are released below.
+    MQuickJsRuntime::Shutdown();
```
cleithner-comcast, line 516: Initialize and Shutdown should be matching lifecycle scopes. Right now, we are initializing in SbmdFactory::RegisterDriversFromDirectory and shutting it down in Matter::Stop. That don't make no sense.

cleithner-comcast (changes requested):

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -119,7 +122,6 @@ namespace barton
 
         /**
          * Log current mquickjs memory usage at the given label.
-         * Also updates peak tracking when walkHeap is enabled.
```
tleacmcsa, line 122: intentionally removed?

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: Without inline, every .cpp file that includes this header would produce its own definition of each static member.

We won't have to worry about this when we move away from this being a pure static class later on anyway.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
tleacmcsa: It is also a really big lambda.  Why not use a function?

docker/patches/mquickjs/0003-add-gc-callback-and-root-count.patch
tleacmcsa, line 1: These two hooks look like they could be very expensive. We might want to have another cmake option to enable this hook (can initially be on).

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
tleacmcsa: Why do you feel this is fixed?  It is still public. Please review these new APIs and scope them appropriately.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
```diff
@@ -159,14 +161,76 @@ namespace barton
...
+
+        /**
+         * Record a JS mutex wait duration in milliseconds.
+         * Called from AcquireJsMutex() in SpecBasedMatterDeviceDriver.cpp.
+         */
+        static void RecordMutexWait(double ms);
+
+        /**
+         * Record a JS exception event.
+         * @param phase  "init" or "loading"
```
tleacmcsa: Why only init or loading phases?  No exceptions tracked at runtime?

Mirroring another class API also has that API documentation in two places that need to be kept in sync.  It is a fragile model.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
```diff
@@ -405,4 +445,44 @@ std::chrono::steady_clock::time_point MQuickJsRuntime::GetDeadline()
...
+{
+    return scriptInterruptFired.load(std::memory_order_relaxed);
+}
+
+bool MQuickJsRuntime::IsContextReady()
+{
+    return jsContextReady.load(std::memory_order_acquire);
+}
+
+void MQuickJsRuntime::RecordMutexWait(double ms)
```
tleacmcsa: Rather than mirror the MQuickJsRuntimeMetrics API in this class, you could provide a getter.

core/src/subsystems/matter/Matter.cpp
rchowdcmcsa: Initialize and Shutdown are now both at the Matter lifecycle level.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
rchowdcmcsa: fixed

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.h
```diff
@@ -216,7 +230,10 @@ namespace barton
          * @param args The args object (consumed by the call)
          * @return Parsed result chain, or nullopt on failure
```
tleacmcsa, line 233: docblock needs updating for new arg

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
rchowdcmcsa: per AI, the goal there was just to guard against spurious OS wakeups. But I've gone and reworked the logic to rely on the predicate overload of wait_until to guard against them.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
rchowdcmcsa: done

config/cmake/options.cmake
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDriver.h
```diff
@@ -50,6 +51,15 @@ extern "C" {
 
 namespace barton
 {
+    /**
+     * Extract the clean driver name from a .sbmd.js file path by stripping the double extension.
+     * e.g. "/path/to/door-lock.sbmd.js" → "door-lock"
+     */
+    inline std::string driverStemFromPath(const std::string &filePath)
+    {
+        return std::filesystem::path(filePath).stem().stem().string();
```
tleacmcsa, line 60: since this doesnt change, you might store in a static or something so the processing doesnt happen every call.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
```diff
@@ -49,6 +50,12 @@ using namespace barton;
 
 bool SbmdFactory::RegisterDrivers()
 {
+    if (!drivers.empty())
+    {
+        icDebug("SBMD drivers already registered; skipping");
```
tleacmcsa: I'm curious why this was needed, as it may indicate a logic problem elsewhere.  We should only register once.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
```diff
@@ -188,41 +198,70 @@ void SbmdFactory::RegisterDriversFromDirectory(const std::string &dirPath, bool
                 if (!file.is_open())
                 {
                     icError("Failed to open SBMD driver: %s", entry.path().c_str());
+#ifdef BARTON_CONFIG_SBMD_METRICS
```
tleacmcsa: I see very few places in other files where this #define is checked before calling a metrics API.  What is the strategy for its use?  Should we only protect invocations where parameter gathering prior to call is expensive? Should we protect every call? Should we rely everywhere on the stubbed implementations each metrics class provides? If we don't rely completely on the stubs, do we even need them or is everything compiled out if the feature is not enabled?

I think the idea for the stubs was to have them so that the invocation points would not have to check a build flag in order to have zero impact, but that idea was before we found the sometimes expensive setup that has to happen before they can be called.  Perhaps the stubs are not useful and everything should be protected behind #ifdefs....

There should be a clear strategy for this.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
```diff
@@ -249,6 +299,13 @@ void SbmdFactory::RegisterDriversFromDirectory(const std::string &dirPath, bool
                 // Store the driver for lifetime management
                 drivers.push_back(std::move(sbmdDriver));
 
+#ifdef BARTON_CONFIG_SBMD_METRICS
+                double loadDurationMs = std::chrono::duration<double, std::milli>(loadEnd - loadStart).count();
+                double heapDelta =
+                    static_cast<double>(usageAfter.heap_used) - static_cast<double>(usageBefore.heap_used);
+                metrics.RecordDriverLoadSuccess(loadDurationMs, heapDelta, driverStem.c_str());
```
tleacmcsa, line 274: It is possible that heapDelta could resolve to 89052.99999999999999999999, right?  Do we need to round somewhere?

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -194,8 +193,25 @@ namespace
     private:
         std::optional<ParsedResult> &result;
     };
+
+    // Acquire the JS runtime mutex and record the wait duration as a metrics observation.
+    std::unique_lock<std::mutex> AcquireJsMutex()
```
tleacmcsa: Why isn't this logic in MQuickJsRuntime?  Every user of the mutex, even in other classes, would have to duplicate this. Perhaps there are none now, but this code wouldnt/shouldnt know that.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1198,10 +1230,30 @@ void SpecBasedMatterDeviceDriver::ExecuteRequestCommand(std::forward_list<std::p
     pending.overallDeadline = std::chrono::steady_clock::now() + std::chrono::milliseconds(overallMs);
     pending.deferralDepth = 0;
 
+    if (opCtx != nullptr)
+    {
+        pending.operationCtx = *opCtx;
+    }
+    else if (driver != nullptr)
+    {
+        pending.operationCtx.driverName = driver->GetDriverStem();
```
tleacmcsa: "driver name" already had a meaning (the name specified inside the sbmd file). You added "stem".  If we don't keep the terminology right, it is going to get confused.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1361,12 +1415,16 @@ void SpecBasedMatterDeviceDriver::HandleDeferredCommandResponse(uint64_t pending
     if (std::chrono::steady_clock::now() > pending.overallDeadline)
     {
         icWarn("Deferred operation %" PRIu64 " exceeded overall deadline", pendingId);
+        metrics.RecordDeferredTimeout(
```
tleacmcsa, line 1406: I have seen "Interrupted" used to indicate a timeout, but here its called "timeout" and I'm confused.  Again, consistent terminology may help.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1488,6 +1547,10 @@ void SpecBasedMatterDeviceDriver::ContinueDeferredChain(PendingOperation &pendin
         icError("Deferred operation %" PRIu64 " exceeded max deferral depth (%u)",
                 pendingId,
                 PendingOperation::MAX_DEFERRAL_DEPTH);
+        metrics.RecordDeferredMaxDepth(
```
tleacmcsa: The name of this method does not imply it is an error path.  Perhaps RecordDeferredDepthExceeded?

core/src/subsystems/matter/Matter.cpp
```diff
@@ -179,6 +180,10 @@ namespace
 Matter::Matter() : groupDataProvider(kMaxGroupsPerFabric, kMaxGroupKeysPerFabric)
 {
     MatterDriverFactory::Instance();
+
```
tleacmcsa, line 183: vestigial comment?

tleacmcsa (changes requested):

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
```diff
@@ -78,36 +78,39 @@ namespace barton
...
                     auto lastTickle = tickleSeq.load(std::memory_order_relaxed);
 
-                    while (true)
+                    // Predicate-based wait handles spurious wakeups internally.
+                    // Returns true when the predicate fires (stop requested or activity);
+                    // returns false on timeout.
+                    bool activityOccurred = samplerCv.wait_until(lock, deadline, [&]() {
+                        return samplerShouldStop.load(std::memory_order_relaxed) ||
+                               tickleSeq.load(std::memory_order_relaxed) != lastTickle;
+                    });
```
cleithner-comcast: This loop is much better now. If I could be nitpicky, I would maybe change activityOccurred to:

`bool timeout= !samplerCv.wait_until(lock, deadline, [&]() {
                        return samplerShouldStop.load(std::memory_order_relaxed) ||
                               tickleSeq.load(std::memory_order_relaxed) != lastTickle;
                    });`

And, as far as I can tell, you can just check for the timeout. Both tickle and samplerShouldStop will reoslve in the next loop appropriate so you don't need the return block.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
cleithner-comcast: I'd be surprised if the compiler didn't figure it out for the TLU but I won't pretend to be certain what that magic box does.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -179,25 +179,23 @@ namespace barton
...
-#endif
+            metrics.RecordOutcome(outDriver, outOpType, outResourceId, "stack_overflow");
 
             return std::nullopt;
         }
 
-#ifdef BARTON_CONFIG_SBMD_METRICS
         // Capture pre-call heap state
         JSMemoryUsage usageBefore = {};
         JS_GetMemoryUsage(ctx, &usageBefore, 0);
```
cleithner-comcast: Is this an expensive call? If so it's maybe a good exception to keep the ifdef around it. Or, if we wanted to be very diligent, we patch an option into the mquickjs and set it appropriately at compile-time so no call-site ifdef check is necessary.

cleithner-comcast:
The change requested status is a leftover, I just have an open question about the call into mquickjs.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
```diff
@@ -0,0 +1,155 @@
...
+     * Observability metrics for the mquickjs JS runtime: heap pool health,
+     * GC activity, mutex contention, and JS exceptions.
+     */
+    class MQuickJsRuntimeMetrics
+    {
+    public:
+        MQuickJsRuntimeMetrics();
+
+        /**
+         * Start the background idle heap sampler.
```
kfundecmcsa: would it be good to expand this doc comment a bit? I had to read the .cpp loop to figure out the actual sampling policy — that it only snapshots after a full idle period with no activity, and that a tickle resets the countdown. Probably useful for future readers

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
```diff
@@ -0,0 +1,229 @@
...
+            JSContext *ctx = MQuickJsRuntime::GetSharedContext();
+
+            if (!ctx)
+            {
+                return;
+            }
+
+            JSMemoryUsage usage = {};
+            size_t gcRootCount = JS_GetGCRootCount(ctx);
+            JS_GetMemoryUsage(ctx, &usage, 0);
```
kfundecmcsa: May consider checking the return value of `JS_GetMemoryUsage` here. Looks like since `usage` is 0 initialized, if an error happened in `JS_GetMemoryUsage`, it would record an all-zeros snapshot, rather than just bailing

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
```diff
@@ -0,0 +1,155 @@
...
+
+        /**
+         * Record a JS exception event.
+         * @param phase  "init" or "loading"
+         * @param driver Filename stem, or nullptr to omit the "driver" attribute
+         */
+        void RecordJsException(const char *phase, const char *driver);
+
+        /**
+         * GC instrumentation callback — registered with JS_SetGCCallback with
```
kfundecmcsa, line 108: Might want to add a blurb on the purpose of `GCCallback` - currently says some of the mechanics, but not really what it eoes. Looking at the implementation, it looks like its for recording the GC frequency and per-cycle duration

core/deviceDrivers/matter/sbmd/metrics/SbmdHandlerInvokerMetrics.h
```diff
@@ -0,0 +1,79 @@
...
+     * heap impact, and outcome counts.
+     */
+    class SbmdHandlerInvokerMetrics
+    {
+    public:
+        SbmdHandlerInvokerMetrics();
+
+        /**
+         * Record handler invocation timing and heap impact.
+         * @param resourceId  nullptr to omit the "resource_id" attribute
```
kfundecmcsa, line 50: were the other `@params` omitted deliberately? just pointing it out in case not

I wasn't sure what `opType` was - had to hunt it down to understand. May wanna just point to `OperationContext.opType` for future readers

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
rchowdcmcsa: with the flags param set to 0, it's a cheap O(1) operation.

core/src/subsystems/matter/Matter.cpp
rchowdcmcsa: I don't know what you're referring to here

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: I changed the spots that used "interrupt" (like WasInterrupted()) to use "timeout"/"timed out"

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: Renamed the variable to driverStem be more clear about what it is.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
rchowdcmcsa: It's being calculated from two size_t's that are being cast to doubles, so I don't think it can here. (This casting happens because the subsequent method takes a double.)

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
rchowdcmcsa: We now rely solely on the stubbed implementations. The only call sites where this #define is still checked is where JS_SetGCCallback is called.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
rchowdcmcsa: AI added this bit in a prior iteration. It is ultimately vestigial dead code, so I'm just getting rid of it.

core/deviceDrivers/matter/sbmd/SbmdDriver.h
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.h
rchowdcmcsa: Done.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
rchowdcmcsa: Done.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: There is one path where runtime exceptions could occur that we were not tracking; I've closed that gap.

> Mirroring another class API also has that API documentation in two places that need to be kept in sync. It is a fragile model.

Removed the unnecessary APIs from MQuickJsRuntime.h and provided a getter to the metrics object.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: My bad, "fixed" was the wrong thing to say here. I had added a comment explaining why it was public.

But I've just now gone and removed the API instead.

The other APIs do need to be public.

docker/patches/mquickjs/0003-add-gc-callback-and-root-count.patch
rchowdcmcsa: Added.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
rchowdcmcsa: I changed it to be a function.

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: Nah, got removed by mistake during some of the refactoring. Added it back.

kfundecmcsa:
I'm making my way through the PR - a lot to digest, I'll make another pass soon

core/deviceDrivers/matter/sbmd/metrics/SbmdHandlerInvokerMetrics.h
rchowdcmcsa: nope, missed that. fixed.

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.cpp
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: done.

core/src/subsystems/matter/Matter.cpp
tleacmcsa: // Initialize SBMD observability metrics before the JS runtime starts so
    // exception counters are live for init-phase exceptions.

I dont see where they are initialized here as the comment implies.

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
tleacmcsa: Marking a comment done should be up to the person leaving the comment unless its something trivial.  Here you again said "done" when you did not do what I asked.  This affects trust :-(

cleithner-comcast (dismissed):
I think all of my concerns have been addressed. I'd call it good once others are satsfied.

core/src/subsystems/matter/Matter.cpp
rchowdcmcsa: Oh. I removed that before I saw this comment, so we should be good if you look at the file in its current state.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -175,34 +179,76 @@ namespace barton
...
             icError("stack overflow before handler call");
+            metrics.RecordOutcome(outDriver, outOpType, outResourceId, "stack_overflow");
+
             return std::nullopt;
         }
 
+        // Capture pre-call heap state
+        JSMemoryUsage usageBefore = {};
+        JS_GetMemoryUsage(ctx, &usageBefore, 0);
+        auto callStart = std::chrono::steady_clock::now();
```
kfundecmcsa, line 197: might wanna check the return values from the `JS_GetMemoryUsage` calls like in the other spot

I spotted at least another usage you might want to adjust in SbmdFactory.cpp

This will probably keep popping up down the road, might be good to just make a little helper to get the usage and error check, something like

```cpp
std::optional<JSMemoryUsage> MQuickJsRuntime::GetMemoryUsage(JSContext *ctx, int flags)
{
    if (!ctx)
    {
        return std::nullopt;
    }

    JSMemoryUsage usage = {};

    if (JS_GetMemoryUsage(ctx, &usage, flags) != 0)
    {
        icWarn("JS_GetMemoryUsage failed");
        return std::nullopt;
    }

    return usage;
}
```

then a safe call would be like

```cpp
auto usageBefore = MQuickJsRuntime::GetMemoryUsage(ctx, 0);
```
up to you though

core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.h
rchowdcmcsa: Introduced the *Metrics.h/cpp files in a later patch

core/deviceDrivers/matter/sbmd/metrics/MQuickJsRuntimeMetrics.h
rchowdcmcsa: MQuickJsRuntimeMetrics no longer follows the singleton model

core/deviceDrivers/matter/sbmd/SbmdDriver.h
```diff
@@ -110,6 +120,15 @@ namespace barton
          */
         const std::string &GetName() const;
 
+        /**
+         * Get the clean driver stem for observability metric attributes.
+         * Strips the .sbmd.js double extension from the file path.
```
kfundecmcsa: this function just returns the `driverStem` though, right? It doesn't strip it from the file path, that's handled by `driverStemFromPath`

tleacmcsa (dismissed):

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
```diff
@@ -1198,10 +1216,30 @@ void SpecBasedMatterDeviceDriver::ExecuteRequestCommand(std::forward_list<std::p
     pending.overallDeadline = std::chrono::steady_clock::now() + std::chrono::milliseconds(overallMs);
     pending.deferralDepth = 0;
 
+    if (opCtx != nullptr)
+    {
+        pending.operationCtx = *opCtx;
+    }
+    else if (driver != nullptr)
```
kfundecmcsa, line 1225: Took me a second to be sure this wasn't a bug when both `opCtx` and `driver` are set. I get it now - `*opCtx` wins and its `driverStem` already comes from the same driver, so the `else if` is just the no-opCtx fallback. Might be good to have a short comment saying the branches are deliberately exclusive (and that a non-null opCtx is assumed to already carry driverStem)? Would save the next reader the same double-take.

kfundecmcsa (commented):
Did my best to finish up - got a little burnt out, nice work though man!

core/deviceDrivers/matter/sbmd/SpecBasedMatterDeviceDriver.cpp
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/SbmdDriver.h
rchowdcmcsa: true. fixed.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
rchowdcmcsa: done

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
```diff
@@ -212,18 +211,23 @@ namespace barton
...
 
         // Capture post-call state and record metrics
-        JSMemoryUsage usageAfter = {};
-        JS_GetMemoryUsage(ctx, &usageAfter, 0);
+        auto usageAfter = MQuickJsRuntime::GetMemoryUsage(ctx, 0);
 
         double durationMs =
             std::chrono::duration<double, std::milli>(std::chrono::steady_clock::now() - callStart).count();
-        double heapDelta = static_cast<double>(usageAfter.heap_used) - static_cast<double>(usageBefore.heap_used);
+        double heapDelta = (usageBefore && usageAfter) ? static_cast<double>(usageAfter->heap_used) -
```
tleacmcsa: what is the value of recording an invocation if we dont have heapDelta due to an error getting memory usage from MQuickJs?  Does this really happen?  Using 0.0 for heap delta would skew the results when these errors occur, right?

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
```diff
@@ -263,8 +263,9 @@ void SbmdFactory::RegisterDriversFromDirectory(const std::string &dirPath, bool
                 drivers.push_back(std::move(sbmdDriver));
 
                 double loadDurationMs = std::chrono::duration<double, std::milli>(loadEnd - loadStart).count();
-                double heapDelta =
-                    static_cast<double>(usageAfter.heap_used) - static_cast<double>(usageBefore.heap_used);
+                double heapDelta = (usageBefore && usageAfter) ? static_cast<double>(usageAfter->heap_used) -
+                                                                     static_cast<double>(usageBefore->heap_used)
+                                                               : 0.0;
```
tleacmcsa: same question about heapDelta being 0.0 on errors.

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
rchowdcmcsa: > Does this really happen? 

Extremely unlikely, as it should only happen if the runtime/context pointer is somehow invalid. I still think it's good practice to account for it happening though.

But, yes, it would skew the results and I did not think of that. There is still value in recording an invocation because duration is an independent histogram and it tells us about slow handlers regardless of whether heap measurement succeeded, but what we can do is pass heapDelta to RecordInvocation as an optional and guard the histogram record inside RecordInvocation with a .has_value() check. Does that sound good?

kfundecmcsa (dismissed):

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
tleacmcsa: i think that would be ok.

tleacmcsa: given a choice between wrong data and no data being collected, i choose no data.   we need to be able to trust the data we get.

cleithner-comcast (dismissed):

core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
rchowdcmcsa: Done.

core/deviceDrivers/matter/sbmd/SbmdFactory.cpp
rchowdcmcsa: Done.

tleacmcsa (approved):

PR #247, opened 2026-07-20
fix(matter): Fix OCW to parse node id as hex

cleithner-comcast:
Previously if you opened a commissioning window via providing a node id then it would fail to parse any node id containing hex digits. Since nodes and device ids (equivalent for matter) are always expressed in hex, this was a fairly impactful bug.


core/src/subsystems/matter/matterSubsystem.cpp
```diff
@@ -542,7 +542,11 @@ bool matterSubsystemOpenCommissioningWindow(const char *nodeId, uint16_t timeout
 
     if (nodeId != nullptr)
     {
-        if (!stringToUint64(nodeId, &_nodeId))
+        // Node IDs are represented as bare (non-prefixed) hexadecimal strings throughout the
+        // reference app (see matterCategory.c openCommissioningWindow, which formats via %016llx).
+        // Parse explicitly as base 16 so hex digits (a-f) are accepted; stringToUint64 uses base 0
+        // and would misinterpret a non-0x-prefixed hex string as decimal and reject it.
+        if (!stringToUnsignedNumberWithinRange(nodeId, &_nodeId, 16, 0, UINT64_MAX))
```
tleacmcsa, line 549: This is potentially contract breaking.  Make sure the 2 CLI interfaces to this command plus the aws/CL interface are ok with this.

cleithner-comcast: The CL interface uses thing id, comm translates to device id. The higher level interface (our public api) is device id. device id, as far as I know, has always been a hexidecimal string so I believe this was an implementation bug not a contract break.

tleacmcsa: Where are you seeing that our api says device id? I see node id.  This API, when given a nonzero nodeId, acts on an other Matter device that the controller can access.  It might not be in our database, therefore it might not have a device id.

cleithner-comcast: > The higher level interface (our public api) is device id

```
/**
 * b_core_client_open_commissioning_window
 * @self: the BCoreClient instance.
 * @deviceId: the device id to open the commissioning window for or NULL to open locally on this device. "0" also
 *            indicates local commissioning.
 * @timeoutSeconds: the number of seconds to keep the commissioning window open or 0 to use the default timeout.

 * @brief Open a Matter commissioning window for a remote device or locally on this device.  Upon success, a 11 digit
 * setup code and a textual QR code will be returned for use by a commissioner to commission the device over the
 * operational network.
 *
 * Returns: (transfer full): BCoreCommissioningInfo* - the info about the open commissioning window upon
 *                           success
 * or NULL upon failure.
 */
BCoreCommissioningInfo *
b_core_client_open_commissioning_window(BCoreClient *self, const gchar *deviceId, guint16 timeoutSeconds);
```


The device id is the node id for matter so even if we're talking about nodes on the fabric that are somehow not in the device db, their node ids are hexidecimal strings as far as I know. So I'm not sure what you are concerned about here. The way I even discovered this bug was with a real camera and opening its commissioning window (which was broken).

tleacmcsa: I'm just responding to where you said the higher level API uses device id.  While true it is in b_core_client_open_commissioning_window as that, it is nodeid just about everywhere else:

bool matterSubsystemOpenCommissioningWindow(const char *nodeId,  ...
bool Matter::OpenCommissioningWindow(chip::NodeId nodeId,  ...
bool deviceServiceOpenCommissioningWindow(const char *nodeId,  ...

It is minor, but the distinction in my mind is the difference between a device we manage (paired with), and some remote Matter device (which wouldn't use our "device id" concept).  Yeah, yeah, they are compatible now, but that was almost not the case (we almost prefixed the fabric id in front of node id for our device id).  They were not supposed to be directly mappable.  Deviceid is just an opaque unique handle to us.

Anyway, I totally get the core issue which was you didn't expect to have to use decimal for matter node ids since no other tool did.  If you are 100% certain that a CL call wont send decimal and that its clear in xhDeviceUtil usage (I didnt look), then I'm ok with it.

cleithner-comcast: I hear ya. At this time I can't say with 100% certainty because I haven't done an empirical test but it is something that I can validate for sure while I do camera end to end. I did look at the code paths and from looking over them, it seems they just are using the device id as the argument. If I somehow discover that I am wrong and the callers are doing something goofy to the device id to make the API "work" (again, I don't see such code right now), then I think fixing the client code there would be correct at that point so we would not make any changes to the barton layer.

Which is a long winded way of saying, I think that this change is the correct change to actually make the **public** API behavior match expectations (device id, the advertised parameter, "just works"). Naturally whether we want to change that API's contract to decouple from device id I think is something we can address in the future if it comes up.

tleacmcsa (approved):

rchowdcmcsa (approved):

PR #251, opened 2026-07-22
fix(matter): guard cachedClusterFeatureMaps against concurrent UpdateCachedFeatureMaps

cleithner-comcast:
UpdateCachedFeatureMaps() runs on two threads: the commissioning worker
(via SpecBasedMatterDeviceDriver::AddDevice) and the Matter event loop
(via CacheCallback::OnSubscriptionEstablished). Both reassigned
cachedClusterFeatureMaps without synchronization, while SBMD handler
threads read it via GetCachedClusterFeatureMaps().

Concurrent move-assignment of the same std::map corrupts its internal
red-black tree and orphans a node, which LeakSanitizer reported as a
40-byte leak in door_lock_test::test_sideband_unlock_triggers_barton_update.
Returning the cache by const reference also allowed a reader to observe
the map mid-reassignment (a read/write data race and use-after-free).

Add cachedClusterFeatureMapsMutex to serialize all access. Build the
new map locally (keeping TLV decode outside the lock) and take the lock
only for the move-assign. Change GetCachedClusterFeatureMaps() to return
by value under the lock so callers get a stable snapshot instead of a
reference that can be invalidated concurrently.


core/deviceDrivers/matter/MatterDevice.h
```diff
@@ -156,8 +157,9 @@ namespace barton
          * Get the cached cluster feature maps.
          * @return Map of cluster ID to feature map value.
```
tleacmcsa: document its a copy?

tleacmcsa (dismissed):

core/deviceDrivers/matter/MatterDevice.h
cleithner-comcast: Done.

tleacmcsa (approved):

rchowdcmcsa (approved):

PR #252, opened 2026-07-22
test(integration): add opt-in parallel test execution

cleithner-comcast:
Run the integration suite in parallel via 'py_test.sh --parallel[=N]' (serial by default so individual/interactive runs keep readable, interleaved logs). CI opts in via run_integration_tests.sh. Default workers is min(nproc/4, 64) -- roughly half the physical cores, since each worker drives a multi-threaded Barton plus a matter.js node process and nproc counts hyperthreads.

Supporting changes for concurrent test processes:
- matter: bind ephemeral operational/UDC ports under BARTON_CONFIG_MATTER_USE_RANDOM_PORT so concurrent Barton servers don't collide on CHIP_PORT/CHIP_UDC_PORT (5540/5550).
- matter: barton patch 0003 raises the CHIP commissioner's discovered-node cache 10 -> 128. On the shared mDNS plane every commissioner sees every device's advertisement, so concurrent commissionings overflowed the 10-slot cache ('Insufficient space') and discovery timed out -- capping reliable parallelism at ~4. With 128 no overflow occurs even at 64 workers.
- test env: per-process storage on tmpfs with cleanup on process exit regardless of outcome; sweep the shared compile-time ~/.brtn-ds once at session end.
- pin the fixed-port zhal mock tests (18443/8711) to one xdist worker via xdist_group + loadgroup; strip xdist's @group nodeid suffix in the subprocess-per-test runner.
- raise commission/ready/resource timeouts (5->30, 10->30) to tolerate load.
- py_test.sh: resolve repo root from script location so PYTHONPATH and LD_LIBRARY_PATH point at this worktree, not a stale primary clone.


docker/version
```diff
@@ -1 +1 @@
-2.13
+2.14
```
cleithner-comcast: 2.15 will be claimed by webrtc work, 2.16 needed here.

testing/environment/base_environment_orchestrator.py
```diff
@@ -182,7 +197,7 @@ def _on_status_changed(self, _object, statusEvent: BCore.StatusEvent):
                 self._ready_to_commission = True
                 self._ready_for_devices_condition.notify_all()
 
-    def wait_for_client_to_be_ready(self, timeout=5):
+    def wait_for_client_to_be_ready(self, timeout=30):
```
cleithner-comcast: timeouts should not have been increased, need to be brought back down.

testing/conftest.py
```diff
@@ -55,6 +55,26 @@ def pytest_configure(config):
...
+def pytest_sessionfinish(session, exitstatus):
+    """Sweep the shared, compile-time Matter config dir once, after everything.
+
+    The Matter SDK's PosixConfig ini files always live at the compile-time
+    CHIP_BARTON_CONF_DIR (~/.brtn-ds/matter) and cannot be redirected at runtime,
+    so every test process shares that one directory. Remove it only from the
+    outer, top-level session -- not the per-test subprocess runs, and not the
+    xdist workers -- so it is never deleted while another concurrent test is
+    still using it. Per-process KVS/dynamic storage lives on tmpfs and is cleaned
+    up by each process itself (see base_environment_orchestrator.py).
```
cleithner-comcast: Not sure I want to do this, it could be shared with reference app runs as well.

testing/py_test.sh
```diff
@@ -41,19 +41,36 @@
...
 
+# Resolve the repository root from this script's own location so the wrapper
+# operates on the working tree it lives in (correct for git worktrees) instead
+# of paths baked into the container environment at launch time. Prepend this
+# tree's Python packages and freshly-built libBartonCore so they take precedence
+# over any stale paths inherited from the environment (e.g. a primary clone).
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." &>/dev/null && pwd)"
+export PYTHONPATH="$REPO_ROOT${PYTHONPATH:+:$PYTHONPATH}"
+export LD_LIBRARY_PATH="$REPO_ROOT/build/core${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```
cleithner-comcast, line 66: These need to be re-looked at after a rebase.

cleithner-comcast (commented):
In addition to the below, need to look at how the suite dumps logs when an error occurs. We need sane separators if logs are interleaved so that they can be parsed.

PR #253, opened 2026-07-22
feat(camera): add Matter WebRTC camera driver support

cleithner-comcast:
Add end-to-end support for a Matter WebRTC camera device driver.

- Add the WebRTCTransportRequestor cluster (Offer/Answer/ICECandidates/End
  commands, CurrentSessions attribute) to the Barton Matter library, defined
  in barton-library.zap and declared as an external cluster; the generated
  barton-library.matter is regenerated from the ZAP via the SDK generator.
- Configure a WebRTC requestor node and grant commissioned peers CASE/operate
  access to the requestor cluster via a shared ACL helper (refactored from the
  OTA-requestor-specific helper). Treat ACL entries with no targets as granting
  all clusters, per CHIP ACL semantics.
- Add the SBMD camera driver spec (camera.sbmd.js) with an abstract camera
  session endpoint and a WebRTC signaling endpoint (offer/answer/ICE resources
  and incoming command handlers), using the volatile mode for the event-only
  webrtcError resource.
- Restrict SBMD attribute/event/command handlers to alias binding (remove the
  inline clusterId form). This is a breaking schema change.
- De-version the SBMD schema surface: a single sbmd-spec-schema.json (no
  version in the filename), the validator renamed validate_sbmd_specs.py, and
  version history tracked in schema/CHANGELOG.md; schemaVersion is still
  validated against the schema's expected version. Update the schema, type
  definitions, specs, and docs.
- Add unit tests for the camera WebRTC driver, and extract the shared,
  driver-agnostic SBMD test harness (SbmdDriverTestBase) used by both the
  camera and handler-invoker tests.


scripts/ci/validate_sbmd_v4_specs.py
cleithner-comcast, line 1: Should we rename this file to be version agnostic?

tleacmcsa: Yes, multiple times during development on v4 i told copilot to stop listing the versions. somewhere along the way it started doing it again I guess.

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema.json
kfundecmcsa, line 1: do we want to replace the 4.0 schema? previously we were keeping the older schemas for compatibility's sake - I think I prefer to keep them around as a the schemas evolve

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -109,26 +157,32 @@ SbmdDriver({
...
 
-    aliases: {},
+    aliases: {
+        incomingOffer: {clusterId: CL_WEBRTC_TRANSPORT_REQUESTOR, commandId: CMD_OFFER},
+        incomingAnswer: {clusterId: CL_WEBRTC_TRANSPORT_REQUESTOR, commandId: CMD_ANSWER},
+        incomingIceCandidates: {
+            clusterId: CL_WEBRTC_TRANSPORT_REQUESTOR,
+            commandId: CMD_ICE_CANDIDATES
+        },
+        incomingEnd: {clusterId: CL_WEBRTC_TRANSPORT_REQUESTOR, commandId: CMD_END},
```
kfundecmcsa: minor: `incomingEnd` reads a bit ambiguously - maybe `incomingEndSession`? (along with `handleIncomingEnd`)

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -227,39 +329,74 @@ function executeCreateSession(args)
...
+        try {
+            accepted = Sbmd.Tlv.decode(raw);
+        } catch (e) {
+            accepted = null;
+        }
+    }
+
+    if (Array.isArray(accepted)) {
+        if (accepted.indexOf(CMD_SOLICIT_OFFER) !== -1) {
+            return ROLE_ANSWERER;
```
kfundecmcsa: may consider a single return statement at the end instead of these early returns

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -268,49 +405,48 @@ function executeStream(args)
...
+    var entryPoint = '/' + deviceId + '/ep/webrtc/r/localSdp';
 
-    var metadata = {
-        sessionId: sessionId,
-        protocol: protocol,
-        nextAction: nextAction,
+    // The abstract stream result only names the active protocol and where to begin signaling.
+    // Everything protocol-specific — including the WebRTC negotiation role (read from
+    // ep/webrtc r/negotiationRole) and the ProvideOffer/SolicitOffer/ProvideAnswer commands — stays
+    // on the webrtc endpoint, so this command carries no WebRTC concepts.
```
kfundecmcsa: this appears contradictory - the comment says "carries no WebRTC concepts," but `entryPoint` hardcodes `ep/webrtc/r/localSdp`, which is directly a webrtc concept

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -319,19 +455,598 @@ function executeDestroySession(args)
...
+        ONE_HOUR_SECS
+    );
+
+    if (wasStreaming) {
+        var endSchema = {
+            webRTCSessionID: {tag: 0, type: 'uint16'},
+            reason: {tag: 1, type: 'enum8'}
+        };
+        var endPayload = Sbmd.Tlv.encodeStruct(
+            {webRTCSessionID: webRTCSessionID, reason: 2},
```
kfundecmcsa: `2` looks to be a magic number - would it be good to pull this out into a named constant? like `WEBRTC_END_REASON_USER_HANGUP` or something?

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -319,19 +455,598 @@ function executeDestroySession(args)
...
+        streamUsage: {tag: 2, type: 'enum8'},
+        originatingEndpointID: {tag: 3, type: 'uint16'},
+        videoStreamID: {tag: 4, type: 'uint16'}
+    };
+
+    var tlvBase64 = Sbmd.Tlv.encodeStruct(
+        {
+            webRTCSessionID: null,
+            sdp: ctx.sdp,
+            streamUsage: 3,
```
kfundecmcsa: another magic number, looks like it's used a bit throughout the file, might be worth naming this guy

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -319,19 +455,598 @@ function executeDestroySession(args)
...
+        detail: 'ProvideOffer failed: ' + (args.error ? args.error.message : 'unknown')
+    };
+
+    return Sbmd.result()
+        .dataModel.updateResource(EP_WEBRTC, 'webrtcError', WEBRTC_ERROR_FAILED, metadata)
+        .success();
+}
+
+function executeLocalIceCandidates(args) {
+    var step = 'start';
```
kfundecmcsa: This `step` concept is a bit unusual - could this be debug instrumentation left over from dev? The `DIAG-STEP[...]` output (with name/lineNumber) reads like something added to pin down where a failure was happening, and this is the only handler that wraps its whole body in a try/catch and tracks progress this way.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -319,19 +455,598 @@ function executeDestroySession(args)
...
+        return Sbmd.result().error('Answer command missing SDP');
+    }
+
+    // Store the camera-allocated webRTCSessionID for use by subsequent commands
+    // (ProvideICECandidates, EndSession)
+    var sessionsJson = args.supplements.transientData[TD_SESSIONS];
+    var sessions = parseSessions(sessionsJson);
+
+    if (sessions && webRTCSessionID !== undefined && webRTCSessionID !== null) {
+        for (var id in sessions) {
```
kfundecmcsa: nit: this "find the 'streaming' session" loop is copy-pasted in ~5 handlers. Could pull it into a little `findStreamingSessionId(sessions)` helper next to `parseSessions` to reduce duplication

kfundecmcsa:
made it through the camera spec, will make another pass through the rest next

core/test/src/SbmdCameraWebrtcTest.cpp
```diff
@@ -0,0 +1,1063 @@
...
+
+            return hctx;
+        }
+
+        JSValue EvalFunc(const char *expr) { return JS_Eval(Ctx(), expr, strlen(expr), "<test>", JS_EVAL_RETVAL); }
+
+        /**
+         * Find a resource execute handler by endpoint and resource ID.
+         */
+        JSValue FindResourceHandler(const std::string &endpointId, const std::string &resourceId)
```
kfundecmcsa, line 137: Non-blocking, but flagging for a potential follow on: a chunk of this fixture is generic SBMD-test harness rather than camera-specific. The `updateResource`/`setMetadata`/`deviceServiceSetMetadata` stubs, `g_updateResourceCalls`, `MakeContext`, and the raw `PrefetchSupplements` plumbing are already copy-pasted in `SbmdHandlerInvokerTest.cpp`. The `Find*`/`Invoke*`/`Encode*`/`Expect*` helpers here aren't shared yet, but they're the kind of thing the next driver test will want too.

Might be worth pulling both into a shared `SbmdDriverTestBase` down the line

core/test/src/SbmdCameraWebrtcTest.cpp
```diff
@@ -0,0 +1,1063 @@
...
+    {
+        std::string sessions = R"({"1":{"state":"streaming","protocol":"webrtc","webRTCSessionID":42}})";
+        ExpectErrorContains(InvokeExecuteHandler("webrtc", "localIceCandidates", R"({"not":"array"})", sessions),
+                            "JSON array");
+    }
+
+    TEST_F(SbmdCameraWebrtcTest, ExecuteOfferIceCandidatesTlvHasCorrectFields)
+    {
+        // ProvideICECandidates (0x0553 cmd 0x05): WebRTCSessionID(0), ICECandidates(1)
+        std::string sessions = R"({"1":{"state":"streaming","protocol":"webrtc","webRTCSessionID":42}})";
```
kfundecmcsa: nit: this `sessions` JSON literal is repeated ~24 times across the file, basically the same 3-4 shapes with different id/state/`webRTCSessionID`. A little helper like `SessionsJson(id, state, sessionId)` would cut the repetition and keep the schema spelled out in one place. Non-blocking.

core/test/src/SbmdCameraWebrtcTest.cpp
```diff
@@ -0,0 +1,1063 @@
...
+        int32_t arrLen = 0;
+        JS_ToInt32(Ctx(), &arrLen, len);
+        EXPECT_GE(arrLen, 1) << "ICECandidates array must have at least one entry";
+    }
+
+    TEST_F(SbmdCameraWebrtcTest, ReproLiveIceCandidates)
+    {
+        std::string sessions = R"({"10":{"state":"streaming","protocol":"webrtc","webRTCSessionID":10}})";
+        std::string input =
+            R"(["candidate:1 1 UDP 2015363327 172.29.0.2 44332 typ host","candidate:2 1 TCP 1015021823 172.29.0.2 9 typ host tcptype active","candidate:3 1 TCP 1010827519 172.29.0.2 51837 typ host tcptype passive","candidate:4 1 UDP 2015363583 fd00:c93:eb1::2 43089 typ host","candidate:5 1 TCP 1015022079 fd00:c93:eb1::2 9 typ host tcptype active","candidate:6 1 TCP 1010827775 fd00:c93:eb1::2 34401 typ host tcptype passive","candidate:7 1 UDP 2015363839 fe80::a4a6:7bff:fe56:592e 50953 typ host","candidate:8 1 TCP 1015022335 fe80::a4a6:7bff:fe56:592e 9 typ host tcptype active","candidate:9 1 TCP 1010828031 fe80::a4a6:7bff:fe56:592e 60109 typ host tcptype passive",""])";
```
kfundecmcsa: nit: this ICE-candidates string is a bit hard to read as one long line. Since C++ concatenates adjacent string literals, you could put each candidate on its own line and it compiles to the exact same value:

```cpp
std::string input =
    R"([)"
    R"("candidate:1 1 UDP 2015363327 172.29.0.2 44332 typ host",)"
    R"("candidate:2 1 TCP 1015021823 172.29.0.2 9 typ host tcptype active",)"
    // ...
    R"("candidate:9 1 TCP 1010828031 fe80::a4a6:7bff:fe56:592e 60109 typ host tcptype passive",)"
    R"("")"
    R"(])";
```

Makes each candidate more readable and the trailing empty-string element (which this test is deliberately checking) stands out instead of hiding at the end.

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema.json
tleacmcsa: I kinda feel like we should rename the file to "sbmd-spec-schema.json" and drop the version from the filename.  Use the CHANGELOG.md to record the changes in the schema versions. We should make it easy to find older versions of the schema via git.

core/test/CMakeLists.txt
```diff
@@ -324,6 +324,28 @@ if (BCORE_MATTER)
...
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/SbmdDriver.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/SbmdDispatch.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/SbmdHandlerInvoker.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/SbmdResultExecutor.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/SbmdLoader.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/SbmdJsUtil.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsRuntime.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/SbmdBundleLoader.cpp
+                ${PROJECT_SOURCE_DIR}/core/deviceDrivers/matter/sbmd/mquickjs/MQuickJsStdlib.c
+        LIBS mquickjs gmock BartonCommon::xhLog cjson
```
kfundecmcsa, line 341: I love the namespacing of `BartonCommon::` - the BartonCommon wiring was added while I was on parental leave so this is the first I've really seen it in action. Very cool and intuitive

docs/SBMD.md
```diff
@@ -44,7 +44,7 @@ supported device type adds too much friction to the goal of broad device support
 
 SBMD addresses this by using textual specification files that map between Matter
 types and Barton resources, enabling new device type support without rebuilding or
-redeploying firmware. Each driver is a single `.sbmd.js` file (schema version 4)
+redeploying firmware. Each driver is a single `.sbmd.js` file (schema version 5)
```
kfundecmcsa: should we just drop the schema version number here? else we have to update it with every iteration

scripts/ci/validate_sbmd_v4_specs.py
kfundecmcsa: agreed

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -67,15 +74,11 @@
...
-//
 // Protocol Abstraction
 // --------------------
 // The camera endpoint does not know or care which streaming protocol is in use.
 // The protocol identifier (e.g., "webrtc", "direct") is stored per session and
-// included in sessionStatus metadata so clients can identify the technology.
+// returned by the stream execute so clients can identify the technology.
 // Currently, this driver hardcodes PROTO_WEBRTC for Matter cameras. Other camera
 // technologies would have their own SBMD drivers that create the same abstract
 // camera endpoint with a different protocol identifier.
```
tleacmcsa: The "M" in "SBMD" is for Matter though.  What are we meaning to say here?

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -109,26 +157,32 @@ SbmdDriver({
...
         // Timing
-        ONE_HOUR_SECS: 3600,
+        ONE_HOUR_SECS: 3600
     },
 
-    barton: {
-        deviceClass: 'camera',
-        deviceClassVersion: 1,
-    },
+    barton: {deviceClass: 'camera', deviceClassVersion: 1},
```
tleacmcsa: device class version 1 was for the older stuff... open home cameras, etc.  Make sure we use a reasonable value here and for profile versions.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
```diff
@@ -33,32 +33,39 @@
...
 //   createSession  [execute]  — Allocates a new session and returns a sessionId.
-//   stream         [execute]  — Starts streaming for a given sessionId. Emits a
-//                               sessionStatus event with status "setup" and
-//                               metadata containing the protocol in use and the
-//                               nextAction URI on the protocol-specific endpoint.
+//   stream         [execute]  — Starts streaming for a given sessionId. Returns a
+//                               JSON object { protocol, entryPoint } identifying
+//                               the active protocol and the resource URI on the
+//                               protocol-specific endpoint the client uses next.
 //   takePicture    [execute]  — Captures a snapshot (not yet implemented).
```
tleacmcsa: Is takePicture really a session command?  it doesnt take the transient data.  I know its not implemented yet, but we dont wanna forget, and as specified now it doesnt use session.

tleacmcsa (commented):
I'm good with it after all comments are addressed/closed.

scripts/ci/validate_sbmd_v4_specs.py
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema.json
cleithner-comcast: I think I've addressed this for now. Please nitpick at it. I did not keep 4.x around because sbmd is still so new but I think keeping old versions would be fine if only to support client custom drivers. With that said, I don't feel strongly because a client with a custom driver can just forwardport the schema version they want to use regardless.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Good catch, done.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done

core/test/src/SbmdCameraWebrtcTest.cpp
cleithner-comcast: Went ahead and did the extraction.

core/test/src/SbmdCameraWebrtcTest.cpp
cleithner-comcast: Done

core/test/src/SbmdCameraWebrtcTest.cpp
cleithner-comcast: Done

docs/SBMD.md
cleithner-comcast: Done

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: The info is valid, but should just be "drivers" not "SBMD drivers". Should be fixed.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Done.

core/deviceDrivers/matter/sbmd/specs/camera.sbmd.js
cleithner-comcast: Made it 2, good catch.

core/deviceDrivers/matter/sbmd/schema/sbmd-spec-schema.json
kfundecmcsa: Sounds good to me

core/test/src/SbmdCameraWebrtcTest.cpp
kfundecmcsa: looks great, nice!

kfundecmcsa (approved):

rchowdcmcsa (approved):

tleacmcsa (approved):

PR #254, opened 2026-07-29
app: add camera streaming command to reference app

cleithner-comcast:
Add a 'cs' camera stream command to the reference app that acts as the in-container WebRTC peer to a Matter camera and serves the decoded stream.

- cameraWebrtcClient: webrtcbin peer (host-only ICE), SDP offer/answer, trickle ICE, H.264 passthrough muxed to fragmented MP4, connectivity watchdog, and a keyframe (RTCP PLI) burst worker to start playback quickly.
- cameraMediaServer: a small HTTP server that serves the live fragmented MP4 to a browser via a Media Source Extensions player, replaying a cached init segment to late joiners.
- cameraCategory / cameraDeviceSession: drive the Matter WebRTC signaling handshake and route media to a file (file://) or the HTTP server.
- Timestamp emitOutput/emitError and drain log/command output on a dedicated thread so the prompt stays intact during blocking commands.
- Gate the command behind the BCORE_REFERENCE_CAMERA_SUPPORT CMake option (enabled in the dev Linux platform); add the GStreamer stack to the image and forward the stream port in the devcontainer.


kfundecmcsa:
I recall we agreed a while back that additions to the reference app will be their own commit type as to not affect barton's versioning. As such - I think this commit type would `app`:

```
-   `app:` Changes to the Reference App (`reference/`)
```

https://github.com/rdkcentral/BartonCore/blob/main/CONTRIBUTING.md#commit-message-format

reference/src/barton-core-reference-io.c
```diff
@@ -290,21 +383,20 @@ void barton_core_reference_io_process(bool useLinenoise, processLineCallback cal
...
+    // main loop is blocked inside a synchronous command, so logging threads never stall on a full pipe.
+    asyncContext->logContext = g_main_context_new();
+    asyncContext->logLoop = g_main_loop_new(asyncContext->logContext, FALSE);
 
     // Setup reading from stdin
     g_autoptr(GInputStream) inputStream = NULL;
     if (useLinenoise && linenoiseStartRead(asyncContext))
     {
         linenoiseStartRead(asyncContext);
+        asyncContext->editActive = true;
```
tleacmcsa, line 398: Why is edit active here?

reference/src/cameraCategory.h
```diff
@@ -0,0 +1,28 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//------------------------------ tabstop = 4 ----------------------------------
+
+#pragma once
+
+#include "category.h"
+
+Category *buildCameraCategory(void);
```
tleacmcsa, line 28: We should make these things self-register some day...

tleacmcsa (approved):

reference/src/barton-core-reference-io.c
cleithner-comcast: So this commit introduced a new command, cs, which is a convenience wrapper around manually setting up a local video stream with a camera. This command is a much longer-running, syncronous command compared to existing commands so far (and the command only finishes when the user manually interrupts or as part of an ungraceful teardown). This surfaced issues with the reference app's log draining which was scheduled on the same gmainloop as commands (default thread context). Without going into the specifics, this created deadlocks in the app where logging couldn't drain and commands couldn't be fulfilled. To alleviate this, this commit moves log draining into its own thread and mainloop. However, emitting logs and linenoise interaction still happen on the main loop.

This creates a coordination problem, particularly with the input prompt. This problem existed before but was all handled in the same thread so no coordination needed. To keep linenoise prompt input display and log draining from trampling or misbehaving, this flag was added. As to why it is set here, it's value pairs with linenoiseStartRead(..) calls (which is linenoiseEditStart(...) call under the hood). So, when we direct linenoise to display a prompt and take input (linenoiseEditStart), set editActive to true so that the log draining thread does not trample over the prompt and input.

reference/src/cameraCategory.h
cleithner-comcast: Agreed, created BARTON-401 for it.

reference/src/cameraDeviceSession.c
```diff
@@ -0,0 +1,312 @@
...
+
+    self->handlerId =
+        g_signal_connect(client, B_CORE_CLIENT_SIGNAL_NAME_RESOURCE_UPDATED, G_CALLBACK(onResourceUpdated), self);
+
+    return self;
+}
+
+bool cameraDeviceSessionOpen(CameraDeviceSession *session)
+{
+    if (session == NULL)
```
kfundecmcsa: may consider the `g_return_val_if_fail` for these initial session checks

reference/src/cameraDeviceSession.c
```diff
@@ -0,0 +1,312 @@
...
+    {
+        return false;
+    }
+
+    g_autofree gchar *uri = g_strdup_printf("/%s/ep/camera/r/stream", session->deviceId);
+    g_autofree gchar *result = NULL;
+
+    if (!b_core_client_execute_resource(session->client, uri, session->sessionId, &result))
+    {
+        return false;
```
kfundecmcsa, line 185: would it be helpful to log a warning/error here?

reference/src/cameraDeviceSession.c
```diff
@@ -0,0 +1,312 @@
...
+    {
+        return NULL;
+    }
+
+    // The negotiation role is a WebRTC concept, so it lives on the webrtc endpoint rather than in
+    // the abstract stream result. Read it lazily on first request and cache it for the session.
+    if (session->role == NULL)
+    {
+        g_autofree gchar *uri = g_strdup_printf("/%s/ep/webrtc/r/negotiationRole", session->deviceId);
+        g_autoptr(GError) err = NULL;
```
kfundecmcsa, line 236: looks like this `err` isn't being used, might be good to leverage it for logging

reference/src/cameraMediaServer.c
```diff
@@ -0,0 +1,540 @@
...
+    g_socket_service_stop(self->service);
+    g_main_context_pop_thread_default(self->context);
+
+    return NULL;
+}
+
+CameraMediaServer *cameraMediaServerCreate(const gchar *bindHost, guint16 port)
+{
+    CameraMediaServer *self = g_new0(CameraMediaServer, 1);
+    self->bindHost = g_strdup(bindHost != NULL ? bindHost : "127.0.0.1");
```
kfundecmcsa: `127.0.0.1` is #defined in `cameraCategory.c` as `CAMERA_DEFAULT_SERVE_HOST` - should that be moved to maybe `cameraMediaServer.h` so both these guys have access to it?

reference/src/cameraMediaServer.c
```diff
@@ -0,0 +1,540 @@
...
+
+    g_list_free(self->pendingClients);
+    self->pendingClients = NULL;
+    g_mutex_unlock(&self->mutex);
+
+    g_clear_object(&self->service);
+
+    if (self->loop != NULL)
+    {
+        g_main_loop_unref(self->loop);
```
kfundecmcsa, line 532: is there a reason this unref happens down here rather than the first `if (self->loop != NULL)` check?

reference/src/cameraMediaServer.c
```diff
@@ -0,0 +1,540 @@
...
+
+                g_mutex_unlock(&self->mutex);
+
+                return FALSE;
+            }
+
+            emitOutput("[camera-stream] Viewer connected, awaiting init segment\n");
+
+            // Claim the connection: our pending-list ref keeps it alive until it is promoted.
+            return TRUE;
```
rchowdcmcsa, line 217: Since a viewer connected, shouldn't onViewer be invoked?

reference/src/cameraWebrtcClient.c
```diff
@@ -0,0 +1,946 @@
...
+        {
+            gst_bus_set_sync_handler(bus, NULL, NULL, NULL);
+            gst_object_unref(bus);
+        }
+
+        gst_element_set_state(client->pipeline, GST_STATE_NULL);
+
+        // Block until the NULL state change fully settles so the streaming task and any in-flight
+        // appsink/bus callbacks (which take `client` as userData) have stopped before we free it.
+        gst_element_get_state(client->pipeline, NULL, NULL, GST_CLOCK_TIME_NONE);
```
rchowdcmcsa: Wouldn't it be safer to have a generous but finite timeout here? And then warn or err when it gets hit.

reference/src/barton-core-reference-io.c
tleacmcsa: Got it, thanks!

reference/src/cameraWebrtcClient.c
```diff
@@ -0,0 +1,946 @@
...
+                self->onClosed(self->userData);
+            }
+
+            break;
+
+        default:
+            break;
+    }
+
+    return GST_BUS_PASS;
```
kfundecmcsa: I was curious why we always return a `GST_BUS_PASS` here, so I talked to AI for a bit about. Turns out with no GMainLoop/async watch, PASS just piles messages onto a queue nobody drains until teardown. `GST_BUS_DROP` would free them as they're handled - might be a more appropriate return value here, but up to you really

reference/src/cameraWebrtcClient.c
```diff
@@ -0,0 +1,946 @@
...
+        if (g_strcmp0(gst_sdp_media_get_media(media), "video") != 0)
+        {
+            continue;
+        }
+
+        for (guint b = 0; b < gst_sdp_media_bandwidths_len(media); b++)
+        {
+            const GstSDPBandwidth *bw = gst_sdp_media_get_bandwidth(media, b);
+
+            if (bw != NULL && g_strcmp0(bw->bwtype, "AS") == 0)
```
kfundecmcsa, line 924: nit: might help future readers to spell out that `AS` = "Application Specific" in a comment - I had to go look it up

reference/src/cameraMediaServer.c
cleithner-comcast: It was previously waiting for the first periodic keyframe but now yes onViewer is invoked to request one explicitly.

reference/src/cameraMediaServer.c
cleithner-comcast: Hopefully the new comment explains the method to the madness.

reference/src/cameraWebrtcClient.c
cleithner-comcast: Done.

reference/src/cameraWebrtcClient.c
cleithner-comcast: Done

rchowdcmcsa (approved):

kfundecmcsa (approved):

PR #256, opened 2026-08-04
fix(camera): report the camera's negotiation role, not the client's

cleithner-comcast:
The ep/webrtc negotiationRole resource is the camera's data model, so it now reports the CAMERA's WebRTC role — offerer when the camera generates the offer (SolicitOffer flow), answerer when it answers the client's offer (ProvideOffer flow) — instead of the role the client must take. An intrinsic cameraIsOfferer() predicate drives both the reported role and the (unchanged) Matter flow selection. The reference app inverts the reported value to choose its own role.

The Matter signaling sequence is unchanged. Driver, reference app, the camera unit test (adds NegotiationRoleReportsCameraRole), and the webrtc-signaling- endpoint / camera-stream-reference-command specs are updated to the camera perspective; archives the openspec change.


tleacmcsa (approved):
I'm good with it after addressing copilot's comments.

kfundecmcsa (approved):

PR #262, opened 2026-08-06
chore: Add support for rtk

cleithner-comcast:
Add support in the devcontainer for rtk and mount the user's ~/.claude directory. To use rtk automatically in vs code copilot chat, enable "chat.useClaudeHooks" in your vsc settings and reload (only works on new sessions). This change assumes the user has rtk installed and initialized globally. It does not run rtk init in the project.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -31,7 +31,8 @@
     // NOTE: VSC will copy some files post-config; .ssh/known_hosts, .gitconfig, etc
     "mounts": [
         "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
-        "source=${localEnv:HOME}/.ssh,target=/home/${localEnv:USER}/.ssh,type=bind"
+        "source=${localEnv:HOME}/.ssh,target=/home/${localEnv:USER}/.ssh,type=bind",
+        "source=${localEnv:HOME}/.claude,target=/home/${localEnv:USER}/.claude,type=bind"
     ],
 
     // By setting this value to true, VSC will not run the entrypoint/CMD in the Dockerfile
--- a/docker/Dockerfile.devcontainer
+++ b/docker/Dockerfile.devcontainer
@@ -62,5 +62,5 @@ RUN npm install -g prettier@3.3.3 lint-staged@15.2.10
 # User local installations
 USER $BUILDER_USER
 
-RUN curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh | sh && \
-    echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.bashrc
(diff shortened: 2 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -31,7 +31,8 @@
     // NOTE: VSC will copy some files post-config; .ssh/known_hosts, .gitconfig, etc
     "mounts": [
         "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
-        "source=${localEnv:HOME}/.ssh,target=/home/${localEnv:USER}/.ssh,type=bind"
+        "source=${localEnv:HOME}/.ssh,target=/home/${localEnv:USER}/.ssh,type=bind",
+        "source=${localEnv:HOME}/.claude,target=/home/${localEnv:USER}/.claude,type=bind"
```
tleacmcsa, line 35: what happens if you don't have a ~/.claude directory?

tleacmcsa (dismissed):

kfundecmcsa (dismissed):

docker/Dockerfile.devcontainer
```diff
@@ -58,3 +58,9 @@ RUN npm install -g @fission-ai/openspec@v1.4.1
 # Formatting tooling for the pre-commit hook: Prettier formats the JavaScript SBMD
 # drivers (see .prettierrc.yaml) and lint-staged runs it over staged files.
 RUN npm install -g prettier@3.3.3 lint-staged@15.2.10
+
+# User local installations
+USER $BUILDER_USER
+
+RUN curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh | sh && \
```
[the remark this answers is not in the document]
cleithner-comcast: It's the devcontainer, I want it to periodically be the latest.

.devcontainer/devcontainer.json
cleithner-comcast: The mount creates the directory on both the host and in the container. The directory would be empty.

cleithner-comcast:
This PR focuses on the preferred install path of a user setting up rtk on the host globally for various tools (could be copilot, claude, opencode, gemini, whatever). If we prefer a local-scoped install instead, I can do that. It just means that token saving data will get wiped each container rebuild and gains won't contribute to user-wide data.

.devcontainer/devcontainer.json
tleacmcsa: Hmm, ok.  I thought i saw this pattern create some sort of bogus file in that situation, but I'll trust ya.

tleacmcsa (approved):

.devcontainer/devcontainer.json
cleithner-comcast: I just tried it with a fake directory, which is how I got that behavior. It's still possible for SOMETHING to write to that directory behind the scenes once in the devcontainer, but the initial effect is what I described.

PR #265, opened 2026-08-18
chore: Install project-specific rtk hooks

cleithner-comcast:
Moving from a global hook to a project one. For now, only copilot's hook is installed. Other tools (eg, opencode, gemini, claude) would need to be installed separately. This commit also forward's the host's history db to the devcontainer so commands and estimated savings from commands within the devcontainer can contribute to the host's global user tracking. If a user does not have rtk installed on the host, this will just create the empty directory ~/.local/share/rtk on the host.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -32,7 +32,7 @@
     "mounts": [
         "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
         "source=${localEnv:HOME}/.ssh,target=/home/${localEnv:USER}/.ssh,type=bind",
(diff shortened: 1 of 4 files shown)


cleithner-comcast:
Ah, ran into the old issue of devcontainer attributing intermediate directories to root instead of the user. Fix shortly...

cleithner-comcast:
Okay turns out the intermediate issue isn't really an issue. I did fix permissions on the host if the directory is missing, however.

kfundecmcsa (approved):

rchowdcmcsa (approved):

PR #269, opened 2026-08-18
feat(drivers): add ONVIF SOAP client with WS-UsernameToken auth

cleithner-comcast:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/core/deviceDrivers/onvif/OnvifSoapClient.cpp
+++ b/core/deviceDrivers/onvif/OnvifSoapClient.cpp
@@ -1,0 +1,406 @@
+//------------------------------ tabstop = 4 ----------------------------------
+//
+// If not stated otherwise in this file or this component's LICENSE file the
+// following copyright and licenses apply:
+//
+// Copyright 2026 Comcast Cable Communications Management, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//------------------------------ tabstop = 4 ----------------------------------
+
+#include "OnvifSoapClient.h"
+
+#include "OnvifXml.h"
+
+#include <curl/curl.h>
+#include <glib.h>
+#include <libxml/parser.h>
+#include <libxml/tree.h>
+
+#include <cstring>
+
+namespace barton
+{
+    namespace onvif
+    {
+
+        namespace
+        {
+
+            const char *const NS_SOAP = "http://www.w3.org/2003/05/soap-envelope";
+            const char *const NS_TDS = "http://www.onvif.org/ver10/device/wsdl";
+            const char *const NS_TRT = "http://www.onvif.org/ver10/media/wsdl";
+            const char *const NS_TT = "http://www.onvif.org/ver10/schema";
+            const char *const NS_WSSE =
+                "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd";
+            const char *const NS_WSU =
+                "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd";
+            const char *const PASSWORD_DIGEST_TYPE =
+                "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-username-token-profile-1.0#PasswordDigest";
+            const char *const BASE64_ENCODING_TYPE =
+                "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-soap-message-security-1.0#Base64Binary";
+
+            std::string XmlEscape(const std::string &in)
+            {
+                std::string out;
+                out.reserve(in.size());
+
+                for (char c : in)
+                {
+                    switch (c)
+                    {
+                        case '&':
+                            out += "&amp;";
+                            break;
+                        case '<':
+                            out += "&lt;";
+                            break;
+                        case '>':
+                            out += "&gt;";
+                            break;
+                        case '"':
+                            out += "&quot;";
+                            break;
+                        case '\'':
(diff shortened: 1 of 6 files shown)


rchowdcmcsa (approved):
Just reviewing the code, it looks good apart from a few issues that Copilot AI already flagged.

core/deviceDrivers/onvif/OnvifSoapClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+namespace barton
+{
+    namespace onvif
+    {
+
+        namespace
+        {
+
+            const char *const NS_SOAP = "http://www.w3.org/2003/05/soap-envelope";
```
kfundecmcsa: what does the `NS` prefix for these variables mean?

core/deviceDrivers/onvif/OnvifSoapClient.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+        std::string
+        OnvifComputePasswordDigest(const std::string &nonceRaw, const std::string &created, const std::string &password)
+        {
+            GChecksum *sha1 = g_checksum_new(G_CHECKSUM_SHA1);
+            g_checksum_update(sha1, reinterpret_cast<const guchar *>(nonceRaw.data()), nonceRaw.size());
+            g_checksum_update(sha1, reinterpret_cast<const guchar *>(created.data()), created.size());
+            g_checksum_update(sha1, reinterpret_cast<const guchar *>(password.data()), password.size());
+
+            guint8 digest[20];
```
kfundecmcsa: `20` looks like a magic number, might be good to define this

core/deviceDrivers/onvif/OnvifXml.cpp
```diff
@@ -0,0 +1,132 @@
...
+extern "C" {
+#include <xmlHelper/xmlHelper.h>
+}
+
+namespace barton
+{
+    namespace onvif
+    {
+
+        std::string OnvifXmlEscape(const std::string &in)
```
kfundecmcsa: we already pull in libxml, so you may want to consider using `libxml2`'s `xmlEncodeSpecialChars` rather than crafting our own escape function

core/test/CMakeLists.txt
```diff
@@ -420,3 +420,29 @@ if (BCORE_OBSERVABILITY_BACKEND STREQUAL "memory")
...
+        SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/OnvifSoapClientTest.cpp
+                ${ONVIF_DIR}/OnvifSoapClient.cpp
+                ${ONVIF_DIR}/OnvifXml.cpp
+        LIBS ${ONVIF_TEST_CURL_LINK_LIBRARIES} ${ONVIF_TEST_XML_LINK_LIBRARIES} gmock BartonCommon::xhXmlHelper
+        INCLUDES ${ONVIF_DIR} ${ONVIF_TEST_XML_INCLUDE_DIRS} ${ONVIF_TEST_CURL_INCLUDE_DIRS}
+    )
+
+    if (BUILD_TESTING)
+        bcore_configure_glib()
+    endif()
```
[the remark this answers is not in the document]
kfundecmcsa: agreed with copilot

tleacmcsa (commented):
I reviewed this PR, which introduces a self-contained ONVIF SOAP client (libcurl + libxml2 + glib) with unit tests. This has already been through extensive review rounds (visible in the commit history) that addressed real security issues:

- one-time `curl_global_init` via `std::call_once`
- `tokensOut.clear()` to avoid stale accumulation
- stripping embedded userinfo from stream/snapshot URIs (credential leak)
- `XML_PARSE_NONET` on all parsers (XXE/network-fetch hardening)
- null-safe XML text/attr helpers
- `responseOut.clear()` before each POST (no response concatenation)
- CSPRNG (`getrandom`) instead of glib's non-cryptographic PRNG for the WS-Security nonce
- only emitting the WS-Security header when both username and password are set
- rejecting SOAP Fault/empty responses as failed `GetDeviceInformation` parses

Given all that, the code is in good shape. A few remaining observations for this round:

**Correctness / robustness**
- `OnvifSoapClient.cpp:279` — `CURLOPT_TIMEOUT` bounds the whole transfer, so a slow/unreachable host is bounded, but there's no `CURLOPT_CONNECTTIMEOUT` set explicitly; not a blocker since `CURLOPT_TIMEOUT` covers connect+transfer, just worth confirming that's the intended behavior for camera discovery timeouts.
- No `CURLOPT_SSL_VERIFYPEER/VERIFYHOST` handling for `https://` service URLs. Most ONVIF cameras are plain HTTP, but if this client is ever pointed at an `https` endpoint with a self-signed cert (common for cheap cameras), it will fail closed with default libcurl verification. Worth a comment/TODO if HTTPS cameras are in scope, since there's currently no way to configure this from `OnvifSoapClient`.
- `StripUriCredentials` (OnvifSoapClient.cpp:163) does a naive `find('@')`/`find('/')`; it won't handle IPv6 literal hosts (`rtsp://user:pass@[::1]:554/...`) correctly (the `@` is still before the first `/`, so it should still work, but bracketed-host edge cases weren't tested). Low risk given typical camera URIs, but could use a test case.

**Minor/style**
- `core/test/CMakeLists.txt:766` and `:785` — the outer `if (BCORE_ONVIF AND BUILD_TESTING)` already guards the whole block, so the inner `if (BUILD_TESTING) bcore_configure_glib() endif()` is redundant (always true inside the outer block). Harmless, but could be simplified to a plain call.
- `OnvifSoapClient::GetStreamUri`/`GetSnapshotUri` share almost identical bodies (build request, escape token, post, parse media URI, error message) — could be factored into a private helper taking the operation name/namespace, but this is a nit, not a defect.

**Tests**
- Good coverage: password digest (known vector + determinism), envelope with/without credentials (and asserts the cleartext password never appears — nice), device-info parsing (including the SOAP-Fault-rejection case), profile token parsing (including the empty-profiles failure case), and stream/snapshot URI parsing including the credential-stripping case.
- No test exercises `OnvifSoapClient::Post`/the full `GetDeviceInformation` etc. over real HTTP (reasonable to skip — would need a mock server — but worth noting there's no coverage of the HTTP-status ≥400 error path or transport-error path at all, only the pure parsing/envelope-building functions are tested).

Overall: solid, already hardened through prior review rounds. Nothing blocking; the HTTPS/cert-verification gap and the missing HTTP-status/transport-error test coverage are the two things I'd flag before merge, though neither is critical for an HTTP-only ONVIF deployment.

